PR #7, opened 2024-10-29
feat: add atoma-state with basic init

Cifko:
Add atoma-state, almost identical to what we have in atoma-node. But adapted for postgres.
Add initiliaization of the configs and services (not all, more coming in next PRs).


atoma-proxy/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+  let (shutdown_sender, shutdown_receiver) = watch::channel(false);
+  let (event_subscriber_sender, event_subscriber_receiver) = flume::unbounded();
+  let (state_manager_sender, state_manager_receiver) = flume::unbounded();
+  
+  let sui_subscriber = atoma_sui::SuiEventSubscriber::new(config.sui, event_subscriber_sender, shutdown_receiver);
+
+  // Initialize your StateManager here
+  let state_manager = AtomaStateManager::new_from_url(config.state.database_url, event_subscriber_receiver, state_manager_receiver).await; 
+  if let Err(e) = state_manager {
+    dbg!(e);
```
jorgeantonio21, line 108: Do not use `dbg!` here, use instead `error!`

jorgeantonio21 (approved):
LGTM 

Left a comment to be addressed

PR #10, opened 2024-11-01
feat: acquire stacks on chat completions

Cifko:


atoma-proxy/Cargo.toml
```diff
@@ -5,15 +5,22 @@ edition.workspace = true
...
 tokio = { workspace = true, features = ["full"] }
+tracing-appender.workspace = true
+tracing-subscriber = { workspace = true, features = ["env-filter", "json", "time"] }
 tracing.workspace = true
-tracing-appender = "0.2.3"
-tracing-subscriber = { version = "0.3.18", features = ["env-filter", "json", "time"] }
+sui-keys.workspace = true
+sui-sdk.workspace = true
+blake2 = "0.10.6"
+reqwest = "0.12.9"
```
jorgeantonio21, line 26: Same as above

atoma-proxy/src/main.rs
```diff
@@ -104,21 +114,38 @@ async fn main() {
...
         config.state.database_url,
         event_subscriber_receiver,
         state_manager_receiver,
     )
-    .await;
-    if let Err(e) = state_manager {
-        error!("State manager error {}", e);
-    }
+    .await
+    .unwrap();
```
jorgeantonio21, line 134: Do not unwrap, I understand that in this situation it might be best to panic on state manager initialization, but in that case, I recommend using `.expect` with a proper panicking message.

atoma-proxy/src/main.rs
```diff
@@ -104,21 +114,38 @@ async fn main() {
...
+
+    let shutdown_receiver_clone = shutdown_receiver.clone();
+    let state_manager_handle = tokio::spawn(async move {
+        state_manager.run(shutdown_receiver_clone).await.unwrap();
```
jorgeantonio21, line 138: Same as above

atoma-proxy/src/main.rs
```diff
@@ -104,21 +114,38 @@ async fn main() {
...
+    });
+
+    let sui_subscriber_handle = tokio::spawn(async move {
+        sui_subscriber.run().await.unwrap();
```
jorgeantonio21, line 142: Same as above

atoma-proxy/src/main.rs
```diff
@@ -104,21 +114,38 @@ async fn main() {
...
+    });
+    let server_handle = tokio::spawn(async move {
+        let atoma_sui_client = AtomaSuiClient::new(config.sui.clone()).await.unwrap();
+        let sui = Sui::new(&config.sui).await.unwrap();
```
jorgeantonio21, line 146: Same as above

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+mod config;
+pub use config::AtomaServiceConfig;
+
+use crate::sui::Sui;
+
+const CHAT_COMPLETIONS_PATH: &str = "/v1/chat/completions";
+const NODE_PUBLIC_ADDRESS_REGISTRATION: &str = "/node/registration";
+
+#[derive(Clone)]
+pub struct ProxyState {
```
jorgeantonio21, line 29: Add docs

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    pub _client: Arc<RwLock<AtomaSuiClient>>,
+    pub state_manager_sender: Sender<AtomaAtomaStateManagerEvent>,
+    pub sui: Arc<RwLock<Sui>>,
+}
+
+fn check_auth(headers: &HeaderMap) -> bool {
```
jorgeantonio21, line 35: Is this a work in progress ? I presume so, as checking that `auth` is a hardcoded value is not of value

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        if let Ok(auth) = auth.to_str() {
+            if auth == "Bearer atoma" {
+                return true;
+            }
+        }
+    }
+    false
+}
+
+pub async fn chat_completions_handler(
```
jorgeantonio21, line 46: Please add docs, it helps understanding the logic of the current handler

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    State(state): State<ProxyState>,
+    headers: HeaderMap,
+    Json(payload): Json<Value>,
+) -> Result<Json<Value>, StatusCode> {
+    if !check_auth(&headers) {
+        return Err(StatusCode::UNAUTHORIZED);
+    }
+
+    // Get the model from the payload, so we can find the task.
+    let model = if let Some(model) = payload.get("model") {
```
jorgeantonio21, line 56:
```suggestion
    let model = payload
    .get("model")
    .ok_or(StatusCode::BAD_REQUEST)?
    .as_str()
    .ok_or(StatusCode::BAD_REQUEST)?
    .to_string();
```

Or even:

```suggestion
let model = payload
    .get("model")
    .ok_or(StatusCode::BAD_REQUEST.with_msg("missing model field"))?
    .and_then(|m| m.as_str())
    .ok_or(StatusCode::BAD_REQUEST.with_msg("model must be a string"))?
    .to_string();
```

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        return Err(StatusCode::BAD_REQUEST);
+    };
+
+    let (result_sender, result_receiver) = oneshot::channel();
+
+    state
+        .state_manager_sender
+        .send(AtomaAtomaStateManagerEvent::GetStacksForModel {
+            model: model.clone(),
+            free_compute_units: 1000,
```
jorgeantonio21, line 72: hardcoded ?

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+            error!("Failed to receive GetStacksForModel result: {:?}", err);
+            StatusCode::INTERNAL_SERVER_ERROR
+        })?
+        .map_err(|err| {
+            error!("Failed to get GetStacksForModel result: {:?}", err);
+            StatusCode::INTERNAL_SERVER_ERROR
+        })?;
+
+    let selected_node_id = if stacks.is_empty() {
+        let (result_sender, result_receiver) = oneshot::channel();
```
jorgeantonio21, line 92: I think it would help to have this logic in its standalone function/method

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        .ok_or_else(|| {
+            error!("No node address found for node {}", selected_node_id);
+            StatusCode::NOT_FOUND
+        })?;
+
+    state
+        .sui
+        .write()
+        .await
+        .send_openai_api_request(payload, node_address)
```
jorgeantonio21, line 162: I wonder if we want to have separation of concerns between the actual Sui client logic and the `Client` responsible to make OpenAI api compatible requests to the nodes.

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        .await
+        .map_err(|err| {
+            error!("Failed to send OpenAI API request: {:?}", err);
+            StatusCode::INTERNAL_SERVER_ERROR
+        })
+        .map(|response| Json(response))
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct NodePublicIpAssignment {
```
jorgeantonio21, line 172: I suggest the following:

1. Add docs
2. Change the above logic to different files, to make it easier to search in the codebase, later on.

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    node_small_id: u64,
+    public_address: String,
+}
+
+#[axum::debug_handler]
+pub async fn node_public_address_registration(
```
jorgeantonio21, line 178: Add docs

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    State(state): State<ProxyState>,
+    Json(payload): Json<NodePublicIpAssignment>,
+) -> Result<Json<Value>, StatusCode> {
+    state
+        .state_manager_sender
+        .send(AtomaAtomaStateManagerEvent::UpdateNodePublicAddress {
```
jorgeantonio21, line 184: This is called the first time a node registers, right ? Why then have a `UpdateNodePublicAddress` handler ?

atoma-proxy/src/server/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        })
+        .map_err(|err| {
+            error!("Failed to send UpdateNodePublicAddress event: {:?}", err);
+            StatusCode::INTERNAL_SERVER_ERROR
+        })?;
+
+    Ok(Json(Value::Null))
+}
+
+pub async fn start_server(
```
jorgeantonio21, line 196: Add docs

atoma-proxy/src/sui/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+        SUI_RANDOMNESS_STATE_OBJECT_ID,
+    },
+    wallet_context::WalletContext,
+    SuiClient,
+};
+use tracing::{info, instrument};
+
+const GAS_BUDGET: u64 = 5_000_000; // 0.005 SUI
+
+pub struct Sui {
```
jorgeantonio21, line 26: Add docs

atoma-proxy/src/sui/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+    wallet_ctx: WalletContext,
+    toma_wallet_id: Option<ObjectID>,
+    atoma_package_id: ObjectID,
```
jorgeantonio21, line 29: A potentially better strategy is to have a network `mainnet`, `devnet`, `testnet` and retrieve the right object ids depending on those. But this is good for now.

atoma-proxy/src/sui/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+    atoma_db_id: ObjectID,
+    toma_package_id: ObjectID,
+}
+
+impl Sui {
+    /// Constructor
+    pub async fn new(sui_config: &AtomaSuiConfig) -> Result<Self> {
+        // let keystore = FileBasedKeystore::new(&config.sui.sui_keystore_path().into())
```
jorgeantonio21, line 37: Remove if not needed

atoma-proxy/src/sui/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+            atoma_db_id: sui_config.atoma_db(),
+            toma_package_id: sui_config.toma_package_id(),
+        })
+    }
+
+    #[instrument(level = "info", skip_all, fields(
+      endpoint = "acquire_new_stack_entry",
+      address = %self.wallet_ctx.active_address().unwrap()
+    ))]
+    pub async fn acquire_new_stack_entry(
```
jorgeantonio21, line 61: Add docs

atoma-proxy/src/sui/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+            response.digest
+        );
+        response
+            .events
+            .and_then(|event| {
+                event.data.get(0).and_then(|event| {
+                    event
+                        .parsed_json
+                        .get("selected_node_id")
+                        .and_then(|node_id| node_id.as_i64())
```
jorgeantonio21, line 108: I think this might not work, because the `u64` types in Sui Move, are parsed into `String` when you extract a `SuiJsonValue` object. For this reason, I believe you need to first extract a str and only then parse it into `u64` (I do recommend doing it for `u64` and not `i64`, because of future potential parsing issues).

atoma-proxy/src/sui/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+                None,
+                None,
+            )
+            .await?;
+        coins
+            .into_iter()
+            .max_by_key(|coin| coin.balance)
+            .map(|coin| coin.coin_object_id)
+            .ok_or_else(|| {
+                anyhow::anyhow!(
```
jorgeantonio21, line 224: Maybe add a error tracing here.

cursor.toml
```diff
@@ -0,0 +1,2 @@
+txDigest = "9hH5mDKNegxazEDAXzACauQdiG4KZqR1VTQoBhgVpQzB"
```
jorgeantonio21, line 1: Do we really want to commit this file ? I presume we want to remove it altogether from the repository, to not conflict with previous state history.

atoma-state/src/types.rs
```diff
@@ -75,30 +60,30 @@ pub struct Stack {
     /// Identifier of the selected node for computation
     pub selected_node_id: i64,
     /// Total number of compute units in this stack
-    pub num_compute_units: i64,
+    pub num_compute_units: i32,
     /// Price of the stack (likely in smallest currency unit)
-    pub price: i64,
+    pub price: i32,
     /// Number of compute units already processed
-    pub already_computed_units: i64,
```
jorgeantonio21, line 82: Why ? In this way it is coherent with the contract logic.

atoma-state/src/types.rs
```diff
@@ -75,30 +60,30 @@ pub struct Stack {
     /// Identifier of the selected node for computation
     pub selected_node_id: i64,
     /// Total number of compute units in this stack
-    pub num_compute_units: i64,
+    pub num_compute_units: i32,
```
jorgeantonio21, line 63: Why ? In this way it is coherent with the contract logic.

atoma-state/src/handlers.rs
```diff
@@ -671,6 +671,49 @@ pub(crate) async fn handle_state_manager_event(
                 .update_stack_total_hash(stack_small_id, total_hash)
                 .await?;
         }
+        AtomaAtomaStateManagerEvent::GetStacksForModel {
```
jorgeantonio21, line 674: All looks good, I suggest adding tracing messages, especially on errors. I should do the same for the atoma node logic.

atoma-state/src/state_manager.rs
```diff
@@ -115,9 +114,12 @@ impl AtomaStateManager {
     /// ```
     #[tracing::instrument(level = "trace", skip_all)]
     pub async fn run(self, mut shutdown_signal: Receiver<bool>) -> Result<()> {
+        dbg!("wakanda");
```
jorgeantonio21, line 117: Please remove, if it is relevant to have a debugging for this method, I suggest using `tracing` and also in the `instrument` declarative macro, adding the name of the function

atoma-state/src/state_manager.rs
```diff
@@ -115,9 +114,12 @@ impl AtomaStateManager {
...
         loop {
             tokio::select! {
                 atoma_event = self.event_subscriber_receiver.recv_async() => {
+                    dbg!("wakanda2");
```
jorgeantonio21, line 121: Remove

atoma-state/src/state_manager.rs
```diff
@@ -115,9 +114,12 @@ impl AtomaStateManager {
...
+                    dbg!(&atoma_event);
```
jorgeantonio21, line 122: Remove

atoma-state/src/state_manager.rs
```diff
@@ -230,13 +232,39 @@ impl AtomaState {
...
     pub async fn get_task_by_small_id(&self, task_small_id: i64) -> Result<Task> {
-        let task = sqlx::query("SELECT * FROM tasks WHERE task_small_id = ?")
+        let task = sqlx::query("SELECT * FROM tasks WHERE task_small_id = $1")
             .bind(task_small_id)
             .fetch_one(&self.db)
             .await?;
         Ok(Task::from_row(&task)?)
     }
 
+    pub async fn get_stacks_for_model(&self, model: &str, free_units: i32) -> Result<Vec<Stack>> {
```
jorgeantonio21, line 242: Add docs please, and `#[instrument(...)]`

atoma-state/src/state_manager.rs
```diff
@@ -230,13 +232,39 @@ impl AtomaState {
...
+        let tasks = sqlx::query(
+            "SELECT stacks.* FROM stacks
+                INNER JOIN tasks ON tasks.task_small_id = stacks.task_small_id 
+                WHERE tasks.model_name = $1 AND stacks.num_compute_units - stacks.already_computed_units >= $2")
```
jorgeantonio21, line 246: I believe the `model_name` is not sufficient to retrieve the right stacks, as we also care about the `security level`, and other parameters. But for now, this looks okay. I suggest adding an issue, for better filtering, though

atoma-state/src/state_manager.rs
```diff
@@ -230,13 +232,39 @@ impl AtomaState {
...
+        .bind(model)
+        .bind(free_units)
+        .fetch_all(&self.db).await?;
+        tasks
+            .into_iter()
+            .map(|task| Stack::from_row(&task).map_err(AtomaStateManagerError::from))
+            .collect()
+    }
+
+    pub async fn get_tasks_for_model(&self, model: &str) -> Result<Vec<Task>> {
```
jorgeantonio21, line 256: Add docs and `#[instrument(...)]`

atoma-state/src/state_manager.rs
```diff
@@ -2099,6 +2119,84 @@ impl AtomaState {
...
+    /// # Example
+    ///
+    /// ```rust,ignore
+    /// use atoma_node::atoma_state::AtomaStateManager;
+    ///
+    /// async fn update_address(state_manager: &AtomaStateManager, small_id: i64, address: String) -> Result<(), AtomaStateManagerError> {
+    ///    state_manager.update_node_public_address(small_id, address).await
+    /// }
+    /// ```
+    pub async fn update_node_public_address(&self, small_id: i64, address: String) -> Result<()> {
```
jorgeantonio21, line 2150: Add `#[instrument(...)]` pls

atoma-state/src/state_manager.rs
```diff
@@ -2099,6 +2119,84 @@ impl AtomaState {
...
+    /// # Example
+    ///
+    /// ```rust,ignore
+    /// use atoma_node::atoma_state::AtomaStateManager;
+    ///
+    /// async fn get_address(state_manager: &AtomaStateManager, small_id: i64) -> Result<Option<String>, AtomaStateManagerError> {
+    ///    state_manager.get_node_public_address(small_id).await
+    /// }    
+    /// ```
+    pub async fn get_node_public_address(&self, small_id: i64) -> Result<Option<String>> {
```
jorgeantonio21, line 2191: Add `#[instrument(...)]` pls

atoma-state/src/state_manager.rs
```diff
@@ -2154,19 +2252,15 @@ pub(crate) mod queries {
     pub(crate) async fn create_tasks(db: &PgPool) -> Result<()> {
         sqlx::query(
             "CREATE TABLE IF NOT EXISTS tasks (
-                task_small_id INTEGER PRIMARY KEY,
+                task_small_id BIGINT PRIMARY KEY,
                 task_id TEXT UNIQUE NOT NULL,
-                role INTEGER NOT NULL,
+                role BIGINT NOT NULL,
```
jorgeantonio21, line 2257: Isn't the role a `u16` after all ?

atoma-state/src/state_manager.rs
```diff
@@ -2233,25 +2327,28 @@ pub(crate) mod queries {
...
             );").execute(db).await?;
-        sqlx::query("CREATE INDEX IF NOT EXISTS idx_stacks_owner_address ON stacks (owner_address);")
-            .execute(db)
-            .await?;
-        sqlx::query("CREATE INDEX IF NOT EXISTS idx_stacks_stack_small_id ON stacks (stack_small_id);")
+        sqlx::query("CREATE INDEX IF NOT EXISTS idx_stacks_owner_address ON stacks (owner);")
             .execute(db)
             .await?;
+        sqlx::query(
+            "CREATE INDEX IF NOT EXISTS idx_stacks_stack_small_id ON stacks (stack_small_id);",
```
jorgeantonio21, line 2348: Don't we want to create an index for model dependent retrievals, as well ?

atoma-state/src/state_manager.rs
```diff
@@ -2341,6 +2438,51 @@ pub(crate) mod queries {
...
+    /// ```rust,ignore
+    /// use sqlx::PgPool;
+    /// use atoma_node::atoma_state::queries;
+    ///
+    /// async fn setup_database(pool: &PgPool) -> Result<(), Box<dyn std::error::Error>> {
+    ///     queries::create_nodes_public_addresses(pool).await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    pub(crate) async fn create_nodes_public_addresses(db: &PgPool) -> Result<()> {
```
jorgeantonio21, line 2474: Add `#[instrument(...)]` pls

jorgeantonio21 (changes requested):
Overall it looks good. I left a few minor comments that need to be addressed before we merge this one.

atoma-proxy/src/sui/mod.rs
Cifko: yeah, you are right

atoma-proxy/src/server/mod.rs
Cifko: since this proxy will manage only single client, we can just use password that can be stored in the config.

atoma-proxy/src/server/mod.rs
Cifko: because the address can change, and then the node have to update it.
Renamed to `upsert` that's more appropriate naming

atoma-proxy/src/server/mod.rs
Cifko: I've extracted the send part, and changed it to sui signature only

atoma-proxy/src/server/mod.rs
Cifko: This is the issue I have, what should be this value? It's not like the openai request has this. It can have the max output tokens. But that doesn't count input. And if it doesn't have the max output, then it depends on the config setting of the node.

atoma-proxy/src/server/config.rs
```diff
@@ -14,6 +14,11 @@ pub struct AtomaServiceConfig {
     ///
     /// This field specifies the address and port on which the Atoma Proxy Server will bind.
     pub service_bind_address: String,
+
+    /// Password for the Atoma Proxy Service.
+    ///
+    /// This password is used to authenticate requests to the Atoma Proxy Service.
+    pub password: String,
```
jorgeantonio21, line 21: Hum, why do we need this ?

atoma-proxy/src/server/mod.rs
```diff
@@ -1,70 +1,317 @@
...
+///
+/// ```rust
+/// let mut headers = HeaderMap::new();
+/// headers.insert("Authorization", "Bearer password".parse().unwrap());
+/// let password = "password";
+///
+/// assert_eq!(check_auth(password, &headers), true);
+/// assert_eq!(check_auth("wrong_password", &headers), false);
+/// ```
+
```
jorgeantonio21, line 79: add `#[instrument]` and remove the space

atoma-proxy/src/server/mod.rs
```diff
@@ -1,70 +1,317 @@
...
+/// * `payload`: The payload of the request.
+///
+/// # Returns
+///
+/// Returns the response from the OpenAI API or an error status code.
+///
+/// # Errors
+///
+/// Returns an error status code if the authentication fails, the model is not found, no tasks are found for the model, no node address is found, or an internal server error occurs.
 pub async fn chat_completions_handler(
```
jorgeantonio21, line 116: add `#[instrument]`

atoma-proxy/src/server/mod.rs
jorgeantonio21: You have the value by computing the number of tokens in the request message, and adding the `max_tokens` in the request.

jorgeantonio21 (commented):
Left a few more comments,

but LGTM

atoma-proxy/src/server/config.rs
Cifko: We need some kind of autheticantion with the atoma-proxy. Since every single proxy will manage single client, it doesn't make sense to do some kind of JWT token auth. Simple password is enough. If we are worried that someone will read this config, we can store hash of the password and compare that.

jorgeantonio21 (approved):
LGTM

PR #18, opened 2024-11-06
feat: add proxy deployment

jorgeantonio21:


atoma-state/src/state_manager.rs
```diff
@@ -75,10 +51,6 @@ impl AtomaStateManager {
         event_subscriber_receiver: FlumeReceiver<AtomaEvent>,
         state_manager_receiver: FlumeReceiver<AtomaAtomaStateManagerEvent>,
     ) -> Result<Self> {
-        let (db_url, db_name) = database_url
```
Cifko, line 78: by removing this, the DB has to exists beforehand. For anyone not using docker, but their own postgres instance has to be aware of that.

Cifko (approved):

Cifko:
And please add `.dockerignore` as well

PR #21, opened 2024-11-08
feat: add tokenizers and estimate token use

Cifko:
Add tokenizers to the proxy for tokens estimation. Update the stacks with this.
During stack creation we have the event so we can update the DB right away.

### DIFF
--- a/atoma-proxy/src/server/mod.rs
+++ b/atoma-proxy/src/server/mod.rs
@@ -12,6 +12,7 @@ use flume::Sender;
 use reqwest::header::AUTHORIZATION;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
+use tokenizers::Tokenizer;
 use tokio::{
     net::TcpListener,
     sync::{oneshot, RwLock},
@@ -26,6 +27,8 @@ use crate::sui::Sui;
 const CHAT_COMPLETIONS_PATH: &str = "/v1/chat/completions";
 const MODELS_PATH: &str = "/v1/models";
 const NODE_PUBLIC_ADDRESS_REGISTRATION: &str = "/node/registration";
+const STACK_ENTRY_COMPUTE_UNITS: u64 = 1000;
+const STACK_ENTRY_PRICE: u64 = 100;
 
 /// Represents the shared state of the application.
 ///
@@ -52,6 +55,21 @@ pub struct ProxyState {
     ///
     /// This password is used to authenticate requests to the atoma proxy service.
     pub password: String,
+
+    /// Tokenizer used for processing text input.
+    ///
+    /// The tokenizer is responsible for breaking down text input into
+    /// manageable tokens, which are then used in various natural language
+    /// processing tasks.
+    pub tokenizers: Arc<Vec<Arc<Tokenizer>>>,
+
+    /// List of available AI models.
+    ///
+    /// This list contains the names or identifiers of AI models that
+    /// the application can use for inference tasks. It allows the
+    /// application to dynamically select and switch between different
+    /// models as needed.
+    pub models: Arc<Vec<String>>,
 }
 
 /// Checks the authentication of the request.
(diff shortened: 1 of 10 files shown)


atoma-proxy/src/server/mod.rs
```diff
@@ -26,6 +27,8 @@ use crate::sui::Sui;
 const CHAT_COMPLETIONS_PATH: &str = "/v1/chat/completions";
 const MODELS_PATH: &str = "/v1/models";
 const NODE_PUBLIC_ADDRESS_REGISTRATION: &str = "/node/registration";
+const STACK_ENTRY_COMPUTE_UNITS: u64 = 1000;
```
jorgeantonio21, line 30: Can you add comments for all these constants ? The ones above, this one, and the ones below

atoma-proxy/src/server/mod.rs
```diff
@@ -250,18 +336,40 @@ async fn get_selected_node(
             error!("No tasks found for model {}", model);
             return Err(StatusCode::NOT_FOUND);
         }
-        Ok(sui
+        // TODO: What should be the default values for the stack entry/price?
+        if total_tokens > STACK_ENTRY_COMPUTE_UNITS {
```
jorgeantonio21, line 340: Why do we have this check ? Total tokens can be as large as possible, if a node can process it.

atoma-proxy/src/server/mod.rs
```diff
@@ -250,18 +336,40 @@ async fn get_selected_node(
...
+            return Err(StatusCode::BAD_REQUEST);
+        }
+        let event = sui
             .write()
             .await
-            .acquire_new_stack_entry(tasks[0].task_small_id as u64, 1000, 100)
+            .acquire_new_stack_entry(
+                tasks[0].task_small_id as u64,
+                STACK_ENTRY_COMPUTE_UNITS,
+                STACK_ENTRY_PRICE,
```
jorgeantonio21, line 350: why is this hardcoded ?

atoma-proxy/src/server/mod.rs
```diff
@@ -250,18 +336,40 @@ async fn get_selected_node(
...
+            error!("Total tokens exceeds maximum limit of {STACK_ENTRY_COMPUTE_UNITS}");
+            return Err(StatusCode::BAD_REQUEST);
+        }
+        let event = sui
             .write()
             .await
-            .acquire_new_stack_entry(tasks[0].task_small_id as u64, 1000, 100)
+            .acquire_new_stack_entry(
+                tasks[0].task_small_id as u64,
+                STACK_ENTRY_COMPUTE_UNITS,
```
jorgeantonio21, line 349: why is this hardcoded ?

atoma-proxy/src/server/mod.rs
```diff
@@ -278,36 +386,40 @@ pub struct NodePublicAddressAssignment {
...
+              "id": model,
+              "object": "model",
+              "created": 1730930595,
+              "owned_by": "vllm",
+              "root": model,
+              "parent": null,
+              "max_model_len": 2048,
+              "permission": [
+                {
+                  "id": format!("modelperm-{}", model),
```
jorgeantonio21, line 406: Where is the metadata below coming from ?

atoma-proxy/src/server/mod.rs
```diff
@@ -278,36 +386,40 @@ pub struct NodePublicAddressAssignment {
...
-        ]
+        "data": state
+        .models
+        .iter()
+        .map(|model| {
+            json!({
+              "id": model,
+              "object": "model",
+              "created": 1730930595,
+              "owned_by": "vllm",
```
jorgeantonio21, line 400: why `vllm` ?

atoma-state/src/handlers.rs
```diff
@@ -31,11 +32,9 @@ pub async fn handle_atoma_event(
             handle_node_task_unsubscription_event(state_manager, event).await
         }
         AtomaEvent::StackCreatedEvent(event) => {
-            if event.owner == state_manager.state.owner {
-                handle_stack_created_event(state_manager, event).await
-            } else {
-                Ok(())
-            }
+            // Don't handle creation here. It's handled when the stack is created right away.
```
jorgeantonio21, line 35: Makes sense, please change it to:

`NOTE: Don't handle creation here. It's handled when the stack is created right away.`

atoma-state/src/state_manager.rs
```diff
@@ -260,12 +255,23 @@ impl AtomaState {
     pub async fn get_stacks_for_model(&self, model: &str, free_units: i64) -> Result<Vec<Stack>> {
```
jorgeantonio21, line 255: Can we add unit tests for the update part ?

jorgeantonio21 (commented):
LGTM, 

left a few minor comments to be addressed.

atoma-proxy/src/server/mod.rs
Cifko: Added todos

atoma-proxy/src/server/mod.rs
Cifko: it should not be, explained above.

atoma-proxy/src/server/mod.rs
Cifko: same as above

atoma-proxy/src/server/mod.rs
Cifko: This is what vllm returned and I just used it. I have to check what this structure should/can.
Added todo.

atoma-proxy/src/server/mod.rs
Cifko: changed to `atoma` :-)

atoma-state/src/handlers.rs
Cifko: done

atoma-state/src/state_manager.rs
Cifko: The problem is that for testing you will need to run postgres. That's why there are no tests in this file.

PR #25, opened 2024-11-08
feat: add the ability to override config with env var

Cifko:
The config of `atoma_service` and `atoma_state` can now be overwritten with env var.
E.g. `ATOMA_STATE__DATABASE_URL` will overwrite
```
[atoma_state]
database_url=""
```

`ATOMA_SERVICE__HF_TOKEN` will overwrite
```
[atoma_service]
hf_token = ""
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,9 +118,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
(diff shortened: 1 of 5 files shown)


jorgeantonio21 (approved):
LGTM

PR #26, opened 2024-11-08
feat: update the estimate to real tokens in stacks

Cifko:
Take the response from the endpoint and update the real usage tokens in stacks table.


Cifko:
This is solved in #23

PR #27, opened 2024-11-12
feat: fill the max_tokens if not present

Cifko:
Fill the max tokens when not present.

### DIFF
--- a/atoma-proxy/src/server/chat_completions.rs
+++ b/atoma-proxy/src/server/chat_completions.rs
@@ -445,7 +445,7 @@ async fn authenticate_and_extract(
     let max_tokens = payload
         .get("max_tokens")
         .and_then(|max_tokens| max_tokens.as_u64())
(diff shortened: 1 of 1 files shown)


atoma-proxy/src/server/mod.rs
```diff
@@ -156,7 +156,7 @@ pub async fn chat_completions_handler(
     let max_tokens = payload
         .get("max_tokens")
         .and_then(|max_tokens| max_tokens.as_u64())
-        .ok_or(StatusCode::BAD_REQUEST)?;
+        .unwrap_or(STACK_ENTRY_COMPUTE_UNITS);
```
jorgeantonio21, line 159: where is this variable comes from ?

jorgeantonio21 (changes requested):

atoma-proxy/src/server/mod.rs
Cifko: from the const, that's the variable we currently create stacks with

PR #31, opened 2024-11-15
fix: proper shutdown/failure

Cifko:
now there is global shutdown
- it's triggered when one of the task fails, or ctrl_c is pressed
- it will properly shutdown other tasks


jorgeantonio21 (approved):
LGTM

PR #35, opened 2024-11-23
feat: add transaction digest to request header when a new stack is acquired

jorgeantonio21:
This functionality is important to allow the node to query the Sui blockchain, if the node's event subscribed has not yet catch the transaction's events (which can happen if sending the request is faster than directly reading the blockchain).


jorgeantonio21:
Addresses #36

Cifko (approved):

PR #37, opened 2024-11-26
feat: check the signature on updating public address of the node

Cifko:
Check the signature when calling update public address for a node.
- check if the signature is valid
- check if the signature is for the same small id that's in the request

### DIFF
--- a/atoma-proxy/src/server/http_server.rs
+++ b/atoma-proxy/src/server/http_server.rs
@@ -1,19 +1,29 @@
+use std::str::FromStr;
 use std::sync::Arc;
 
 use anyhow::Result;
 use atoma_state::types::AtomaAtomaStateManagerEvent;
+use atoma_utils::verify_signature;
+use axum::body::Body;
+use axum::extract::Request;
 use axum::http::StatusCode;
+use axum::http::{HeaderMap, StatusCode};
 use axum::middleware::from_fn_with_state;
 use axum::{
     extract::State,
     routing::{get, post},
     Json, Router,
 };
+use blake2::digest::consts::U32;
+use blake2::digest::generic_array::GenericArray;
+use blake2::{Blake2b, Digest};
 use flume::Sender;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
+use sui_sdk::types::base_types::SuiAddress;
+use sui_sdk::types::crypto::{PublicKey, Signature, SuiSignature};
 use tokenizers::Tokenizer;
-use tokio::sync::watch;
+use tokio::sync::{oneshot, watch};
 use tokio::{net::TcpListener, sync::RwLock};
 use tower::ServiceBuilder;
 use tracing::{error, instrument};
@@ -53,6 +63,9 @@ pub const MODELS_PATH: &str = "/v1/models";
 /// This endpoint is used to register or update the public address of a node
 /// in the system, ensuring that the system has the correct address for routing requests.
 pub const NODE_PUBLIC_ADDRESS_REGISTRATION_PATH: &str = "/node/registration";
+
+/// Body size limit for signature verification (contains the body size of the request)
+const MAX_BODY_SIZE: usize = 1024 * 1024; // 1MB
 
 /// Represents the shared state of the application.
 ///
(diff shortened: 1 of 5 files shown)


atoma-proxy/src/server/http_server.rs
```diff
@@ -263,8 +275,90 @@ pub(crate) struct NodePublicAddressRegistrationOpenApi;
...
+        .map_err(|err| {
+            error!("Failed to get node Sui address: {:?}", err);
+            StatusCode::INTERNAL_SERVER_ERROR
+        })?
+        .ok_or_else(|| {
+            error!("Node Sui address not found");
+            StatusCode::NOT_FOUND
+        })?;
+
+    // Check if the small ID in the request matches the Sui address in the signature.
```
jorgeantonio21, line 356: This is not a small id check !

atoma-proxy/src/server/http_server.rs
```diff
@@ -263,8 +275,90 @@ pub(crate) struct NodePublicAddressRegistrationOpenApi;
...
+    if node_sui_address != sui_address.to_string() {
+        error!("Node Sui address does not match the node s");
```
jorgeantonio21, line 358: The error message seems to be incomplete

atoma-proxy/src/server/http_server.rs
```diff
@@ -263,8 +275,90 @@ pub(crate) struct NodePublicAddressRegistrationOpenApi;
...
+        })?;
+    let sui_address = SuiAddress::from(&public_key);
+
+    let mut blake2b_hash = Blake2b::new();
+    blake2b_hash.update(&body_bytes);
+    let body_blake2b_hash: GenericArray<u8, U32> = blake2b_hash.finalize();
+    let body_blake2b_hash_bytes: [u8; 32] = body_blake2b_hash
+        .as_slice()
+        .try_into()
+        .expect("Invalid Blake2b hash length");
```
jorgeantonio21, line 319: So we really want to panic here ?

atoma-state/src/state_manager.rs
```diff
@@ -2160,23 +2162,17 @@ impl AtomaState {
...
-                    (node_small_id, public_address) 
-                    VALUES ($1, $2) 
-                    ON CONFLICT (node_small_id)
-                    DO UPDATE SET public_address = EXCLUDED.public_address",
-        )
-        .bind(small_id)
-        .bind(address)
-        .execute(&self.db)
-        .await?;
+        sqlx::query("UPDATE nodes SET public_address = $2 WHERE node_small_id = $1")
```
jorgeantonio21, line 2165: I do not understand how we can update a node address, isn't this its public key ?

jorgeantonio21 (changes requested):
Looking good, but some changes are required

atoma-proxy/src/server/http_server.rs
Cifko: Changed

atoma-proxy/src/server/http_server.rs
Cifko: no

atoma-state/src/handlers.rs
```diff
@@ -607,6 +607,18 @@ pub(crate) async fn handle_stack_attestation_dispute_event(
     Ok(())
 }
 
+pub(crate) async fn handle_node_registration_event(
```
jorgeantonio21, line 610: add docs and instrument

atoma-state/src/handlers.rs
```diff
@@ -607,6 +607,18 @@ pub(crate) async fn handle_stack_attestation_dispute_event(
...
+    state_manager: &AtomaStateManager,
+    event: NodeRegisteredEvent,
+    address: String,
+) -> Result<()> {
+    state_manager
+        .state
+        .insert_new_node(event.node_small_id.inner as i64, address)
+        .await?;
```
jorgeantonio21, line 618: log potential error here

atoma-proxy/src/server/http_server.rs
```diff
@@ -297,8 +309,91 @@ pub(crate) struct NodePublicAddressRegistrationOpenApi;
 #[instrument(level = "info", skip_all)]
 pub async fn node_public_address_registration(
     State(state): State<ProxyState>,
-    Json(payload): Json<NodePublicAddressAssignment>,
+    headers: HeaderMap,
+    request: Request<Body>,
 ) -> Result<Json<Value>, StatusCode> {
+    let base64_signature = headers
+        .get("X-Signature")
```
jorgeantonio21, line 316: add the right constant

atoma-proxy/src/server/http_server.rs
```diff
@@ -297,8 +309,91 @@ pub(crate) struct NodePublicAddressRegistrationOpenApi;
...
+    let public_key_bytes = signature.public_key_bytes();
+    let public_key =
+        SuiPublicKey::try_from_bytes(signature.scheme(), public_key_bytes).map_err(|e| {
+            error!("Failed to extract public key from bytes, with error: {e}");
+            StatusCode::BAD_REQUEST
+        })?;
+    let sui_address = SuiAddress::from(&public_key);
+
+    let mut blake2b_hash = Blake2b::new();
+    blake2b_hash.update(&body_bytes);
```
jorgeantonio21, line 348: what is the body look like ? can you give an example ?

Also should the node repository have a method to call this endpoint ?

atoma-proxy/src/server/http_server.rs
```diff
@@ -297,8 +309,91 @@ pub(crate) struct NodePublicAddressRegistrationOpenApi;
...
+    let body_blake2b_hash: GenericArray<u8, U32> = blake2b_hash.finalize();
+    let body_blake2b_hash_bytes: [u8; 32] =
```
jorgeantonio21, line 350: replace `32`with proper constant

jorgeantonio21 (commented):
Looking better but still some changes are required.

I do believe we need a p2p network so that we don't need to call this method, as there might be 1000s of proxies and node operators can't possible register to all of them...

Cifko:
> Looking better but still some changes are required.
> 
> I do believe we need a p2p network so that we don't need to call this method, as there might be 1000s of proxies and node operators can't possible register to all of them...

I agree, but currently we can't deploy working network without this. And we don't have p2p implemented. So if we don't plant o hardcode the addresses into the proxy right now, we need this pr to go in.

atoma-state/src/handlers.rs
Cifko: the errors are propagated all the way to error `handle_tasks_results` in `main.rs` We should not double log errors.

atoma-proxy/src/server/http_server.rs
Cifko: yes we should
The body is of type `NodePublicAddressAssignment`
```
pub struct NodePublicAddressAssignment {
    /// Unique small integer identifier for the node
    node_small_id: u64,
    /// The public address of the node
    public_address: String,
}
```

jorgeantonio21 (approved):
LGTM

PR #40, opened 2024-11-28
chore: start confidential compute service and state

jorgeantonio21:


atoma-state/src/state_manager.rs
```diff
@@ -2804,6 +2899,59 @@ pub(crate) mod queries {
...
+    /// }
+    /// ```
+    pub(crate) async fn create_node_public_keys(db: &PgPool) -> Result<()> {
+        sqlx::query(
+            "CREATE TABLE IF NOT EXISTS node_public_keys (
+                node_small_id BIGINT PRIMARY KEY,
+                epoch BIGINT NOT NULL,
+                node_badge_id TEXT NOT NULL,
+                public_key BYTEA NOT NULL,
+                tee_remote_attestation_bytes BYTEA NOT NULL
```
Cifko, line 2947: no foreign keys?

Cifko (approved):
utACK
you could have split this into introducing middleware and confidential in follow up PR

PR #51, opened 2024-12-10
feat: add utoipa to atoma-proxy-service

jorgeantonio21:


atoma-proxy-service/src/proxy_service.rs
```diff
@@ -498,7 +592,7 @@ async fn get_all_subscriptions(
     )
```
francis2tm, line 592: Missing path argument `task_id` in utoipa docs, [see this for reference ](https://github.com/atoma-network/atoma-node/blob/main/atoma-daemon/src/handlers/almost_filled_stacks.rs#L43C1-L45C7)

atoma-proxy-service/src/proxy_service.rs
```diff
@@ -587,7 +700,7 @@ async fn health() -> StatusCode {
...
```
francis2tm, line 700: Missing path argument `node_small_id` in utoipa docs

atoma-proxy-service/src/proxy_service.rs
```diff
@@ -288,7 +346,7 @@ async fn generate_api_token(
...
 )]
 #[instrument(level = "info", skip_all)]
-async fn revoke_api_token(
+pub(crate) async fn revoke_api_token(
     State(proxy_service_state): State<ProxyServiceState>,
     headers: HeaderMap,
     body: Json<RevokeApiTokenRequest>,
```
francis2tm, line 352: This apply to all endpoints. In utoipa docs for this endpoint:
```
(status = OK, description = "Revokes an API token for the user", body = Value),
```
`body = Value`, instead of `body = RevokeApiTokenRequest`

Is it on porpose?

atoma-proxy-service/src/proxy_service.rs
jorgeantonio21: Great ! Thanks for spotting this

atoma-proxy-service/src/proxy_service.rs
jorgeantonio21: Addressed

francis2tm (approved):

PR #62, opened 2024-12-13
feat: validate api token on inference request

Cifko:
Validate the api token on inference request (check if it's in the DB, all revoked tokens are deleted).

### DIFF
--- a/atoma-state/src/state_manager.rs
+++ b/atoma-state/src/state_manager.rs
@@ -2910,14 +2910,11 @@ impl AtomaState {
     /// }
     /// ```
     #[instrument(level = "trace", skip(self))]
-    pub async fn is_api_token_valid(&self, user_id: i64, api_token: &str) -> Result<bool> {
-        let is_valid = sqlx::query(
-            "SELECT EXISTS(SELECT 1 FROM api_tokens WHERE user_id = $1 AND token = $2)",
-        )
-        .bind(user_id)
-        .bind(api_token)
-        .fetch_one(&self.db)
-        .await?;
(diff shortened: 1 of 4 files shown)


atoma-state/src/state_manager.rs
```diff
@@ -2910,14 +2910,11 @@ impl AtomaState {
...
-        let is_valid = sqlx::query(
-            "SELECT EXISTS(SELECT 1 FROM api_tokens WHERE user_id = $1 AND token = $2)",
-        )
-        .bind(user_id)
-        .bind(api_token)
-        .fetch_one(&self.db)
-        .await?;
+    pub async fn is_api_token_valid(&self, api_token: &str) -> Result<bool> {
+        let is_valid = sqlx::query("SELECT EXISTS(SELECT 1 FROM api_tokens WHERE token = $1)")
+            .bind(api_token)
```
jorgeantonio21, line 2915: why don't you need the user_id anymore ?

jorgeantonio21 (approved):
LGTM

jorgeantonio21:
Can we add a feature flag for authentication through bearer tokens only if that feature flag is on ?

atoma-state/src/state_manager.rs
Cifko: I'm relying on the tokens to be not equal, and I don't need any verification with the user_id (it's needed only on the jwt tokens)

Cifko:
> Can we add a feature flag for authentication through bearer tokens only if that feature flag is on ?

I can disable it by feature flags for the openai endpoints, but not endpoints like create api token, etc. because when the jwt is a bearer token it contains information about the user.
I can also disable signature check on the `/node/registration`

PR #63, opened 2024-12-13
feat: quote verification

jorgeantonio21:


atoma-state/src/handlers.rs
```diff
@@ -1067,3 +1069,100 @@ pub(crate) async fn handle_node_key_rotation_event(
...
+            .await
+            .map_err(|e| AtomaStateManagerError::FailedToRetrieveCollateral(format!("{e:?}")))?;
+        let now = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .map_err(|e| AtomaStateManagerError::UnixTimeWentBackwards(e.to_string()))?
+            .as_secs() as u64;
+        let quote = Quote::parse(&quote_bytes)
+            .map_err(|e| AtomaStateManagerError::FailedToParseQuote(format!("{e:?}")))?;
+        match quote.report {
+            Report::SgxEnclave(_) => {
```
dayeol, line 1144: nit: why did you even need to handle it? shouldn't it just fail if it's not TD report?

atoma-state/src/handlers.rs
```diff
@@ -1067,3 +1069,100 @@ pub(crate) async fn handle_node_key_rotation_event(
...
+    /// # Notes
+    ///
+    /// * Uses Intel's PCCS service at a hardcoded URL with a 10-second timeout
+    /// * The `new_public_key` parameter is currently passed through but not used in the verification process
+    /// * This function is specifically for Quote V4 format attestations
+    pub(crate) async fn verify_quote_v4_attestation(
+        quote_bytes: &[u8],
+        new_public_key: &[u8],
+    ) -> Result<()> {
+        let collateral = get_collateral(PCCS_URL, quote_bytes, TIMEOUT)
```
dayeol, line 1134: let's make sure if this is safe from DNS spoofing attacks, e.g., if it checks Intel certificate etc

dayeol (commented):
just a few nits

atoma-state/src/handlers.rs
jorgeantonio21: I presume the current quote could be SGX as well, which is not currently supported by the protocol (given the incompatibility with NVIDIA TEEs). Does it make sense ?

dayeol: My question was - why do you have the logic in the first place instead of just ignore everything else other than TD report. it's ok to keep it though

atoma-state/src/handlers.rs
dayeol: I checked the code and it's probably OK. I think this collateral is very crucial for the overall security, so we may need  some extra integrity checking in the future.

jorgeantonio21: Can we please add the extra integrity checks as GitHub issues in this repository ?

atoma-state/src/state_manager.rs
```diff
@@ -241,35 +241,55 @@ impl AtomaState {
...
-        .await?;
-        tasks
+            RETURNING stacks.*"#,
+        );
+
+        let stacks = sqlx::query(&query)
+            .bind(model)
+            .bind(free_units)
+            .bind(user_id)
+            .fetch_all(&self.db)
```
Cifko, line 288: I just noticed that we have LIMIT 1 so we can change this to fetch_optinal, but it doesn't have to be part of this PR, it's just an observation

atoma-state/src/handlers.rs
```diff
@@ -1105,14 +1105,124 @@ pub(crate) async fn handle_node_key_rotation_event(
         new_public_key,
         tee_remote_attestation_bytes,
     } = event;
+    // TODO: We need to check that the
```
Cifko, line 1108: incomplete `TODO`

atoma-state/src/handlers.rs
```diff
@@ -1105,14 +1105,124 @@ pub(crate) async fn handle_node_key_rotation_event(
...
+            .await
+            .is_ok();
     state_manager
         .state
         .update_node_public_key(
             node_id.inner as i64,
             epoch as i64,
             new_public_key,
             tee_remote_attestation_bytes,
+            is_valid,
```
Cifko, line 1120: The problem I see here, correct me if I'm wrong.
I register TEE node. The node will be selected. And then it rotate to new key which `is_valid` is false. Because the node falsify the key. That way the node can still do the job undetected, but the privacy is gone.
I can do this one time per epoch (current contract conditions it's one time per epoch rotation of key is allowed). It's not much, but still.
Probably once the stack is selected the key should stay the same for the duration of the stack.

Cifko (approved):

atoma-state/src/handlers.rs
jorgeantonio21: It's not a TODO anymore, I remove it :)

atoma-state/src/handlers.rs
```diff
@@ -1067,3 +1069,100 @@ pub(crate) async fn handle_node_key_rotation_event(
...
+mod utils {
+    use super::*;
+
+    use dcap_qvl::collateral::get_collateral;
+    use dcap_qvl::quote::{Quote, Report};
+    use dcap_qvl::verify::verify;
+    use std::time::Duration;
+
+    /// The PCCS URL to obtain collateral for quote verification.
+    const PCCS_URL: &str = "https://pccs.intel.com/sgx/certification/v3/";
```
dayeol: is this correct URL?

atoma-state/src/handlers.rs
jorgeantonio21: The x25519 dh public key is not stored in the stack. If you submit an invalid ra, then the proxy don't select you to process those stacks anymore, and you lose the opportunity to accrue value from fees. Moreover, anyone can now start a dispute with the node. That said, I agree this is not the ideal flow, and in time, the quote verification will be peformed on-chain.

atoma-state/src/handlers.rs
jorgeantonio21: This was removed, yesterday

dayeol (approved):

atoma-state/src/state_manager.rs
Cifko: Because we can send only one to the node anyway. And you want to also update only one with the estimated compute units.

atoma-state/src/handlers.rs
Cifko: Yeah. No new stack will be assigned but the old ones lose privacy. In some cases this can be really bad. Maybe in the future we can store the keys on the stack.

Cifko (approved):

PR #68, opened 2024-12-17
feat: add user_id to stacks

Cifko:
Add `user_id` to `stacks` table.
The stacks that proxy creates are owned by individual users.


jorgeantonio21 (approved):
LGTM

PR #69, opened 2024-12-17
chore: cleanup the leftovers after static password removal

Cifko:
Just remove the missed static proxy password references.


jorgeantonio21 (approved):
LGTM

PR #77, opened 2024-12-19
feat: add logic for node selection in case of confidential compute

jorgeantonio21:


atoma-proxy/src/server/components/openapi.rs
```diff
@@ -39,6 +43,7 @@ pub fn openapi_routes() -> Router {
             (path = CONFIDENTIAL_EMBEDDINGS_PATH, api = ConfidentialEmbeddingsOpenApi, tags = ["Confidential Embeddings"]),
             (path = IMAGE_GENERATIONS_PATH, api = ImageGenerationsOpenApi, tags = ["Images"]),
             (path = CONFIDENTIAL_IMAGE_GENERATIONS_PATH, api = ConfidentialImageGenerationsOpenApi, tags = ["Confidential Images"]),
+            (path = ENCRYPTION_PUBLIC_KEY_ENDPOINT, api = SelectNodePublicKeyOpenApi, tags = ["Node Public Key Selection"])
```
Cifko, line 46: maybe we should group some of them, now every endpoint has its own category (tag)
e.g. confidential would have all confidential stuff, including this endpoint
since it tagging and not a category maybe we can have category with confidential, and category with e.g. image which will also have confidential image

atoma-state/src/state_manager.rs
```diff
@@ -325,47 +341,229 @@ impl AtomaState {
...
+        query.push_str(
+            r#"
             ORDER BY node_subscriptions.price_per_compute_unit 
-            LIMIT 1",
+            LIMIT 1"#,
+        );
+
+        let node_settings = sqlx::query(&query)
+            .bind(model)
+            .bind(false)
```
Cifko, line 431: the 2nd bind doesn have $2 in the query string

atoma-state/src/state_manager.rs
```diff
@@ -325,47 +341,229 @@ impl AtomaState {
...
-                  WHERE valid = true) AS node_subscriptions 
-            ON tasks.task_small_id=node_subscriptions.task_small_id 
+    pub async fn get_cheapest_node_for_model(
+        &self,
+        model: &str,
+        is_confidential: bool,
+    ) -> Result<Option<CheapestNode>> {
+        let mut query = String::from(
+            r#"
+            SELECT tasks.task_small_id, price_per_compute_unit, max_num_compute_units, node_subscriptions.node_small_id
```
Cifko, line 398: I haven't checked but I hope the `price_per_compute_unit` and `max_num_compute_units` are only in one table. Just to be safe can you add the table names as well.

atoma-state/src/state_manager.rs
```diff
@@ -325,47 +341,229 @@ impl AtomaState {
...
-                  FROM node_subscriptions
-                  WHERE valid = true) AS node_subscriptions 
-            ON tasks.task_small_id=node_subscriptions.task_small_id 
+    pub async fn get_cheapest_node_for_model(
+        &self,
+        model: &str,
+        is_confidential: bool,
+    ) -> Result<Option<CheapestNode>> {
+        let mut query = String::from(
+            r#"
```
Cifko, line 397: add a `todo` for the future to double check the performance of this query, since it's triple join it can blow up.

Cifko (approved):

atoma-state/src/state_manager.rs
jorgeantonio21: How come ?

atoma-state/src/state_manager.rs
jorgeantonio21: Done

atoma-state/src/state_manager.rs
jorgeantonio21: Oh good catch !

atoma-proxy/src/server/components/openapi.rs
jorgeantonio21: I added a `Confidential` word to the tag for the public key encryption endpoint

atoma-state/src/state_manager.rs
Cifko: depends how many nodes there are and for how many we buy stacks
but you are right, if we do truncate the settlet stacks it should be fine

Cifko (approved):

PR #80, opened 2024-12-19
feat: key counter for selecting cheapest nodes for confidential compute

jorgeantonio21:


atoma-state/src/state_manager.rs
```diff
@@ -2902,6 +2910,66 @@ impl AtomaState {
...
+    #[instrument(level = "trace", skip_all, fields(%epoch, %key_rotation_counter))]
+    pub async fn insert_new_key_rotation(
+        &self,
+        epoch: i64,
+        key_rotation_counter: i64,
+    ) -> Result<()> {
+        sqlx::query(
+            "INSERT INTO key_rotations (epoch, key_rotation_counter) VALUES ($1, $2)
+            ON CONFLICT (epoch)
+            DO UPDATE SET epoch = EXCLUDED.epoch,
```
Cifko, line 2963: you don't need to set the epoch on conflict on epoch, it's equal already

Cifko (approved):

atoma-state/src/state_manager.rs
jorgeantonio21: Addressed !

PR #81, opened 2024-12-19
feat: add sui address for the user

Cifko:
Add sui_address to the users table.
Add endpoint to store user address. The users needs to send signed message.
The address is derived from the key that signed the message.
The user_id is derived from the access_token (that was also validated).


atoma-auth/src/auth.rs
```diff
@@ -355,6 +369,69 @@ impl Auth {
...
+            error!("Failed to parse signature: {}", e);
+            anyhow::anyhow!("Failed to parse signature {e}")
+        })?;
+        let signature_bytes = signature.signature_bytes();
+        let public_key_bytes = signature.public_key_bytes();
+        let signature_scheme = signature.scheme();
+        let intent_msg = IntentMessage::new(
+            Intent::personal_message(),
+            PersonalMessage {
+                message: "Sign this message to prove you are the owner of this wallet"
```
jorgeantonio21, line 396: Hardcoded messages signign is generally not secure

atoma-state/src/migrations/20241219081058_add_sui_address_to_users.sql
```diff
@@ -0,0 +1,2 @@
+-- Add migration script here
```
jorgeantonio21, line 1: Remove this

atoma-state/src/state_manager.rs
```diff
@@ -3585,6 +3585,43 @@ impl AtomaState {
...
+    ///
+    /// ```rust,ignore
+    /// use atoma_node::atoma_state::AtomaStateManager;
+    ///
+    /// async fn store_sui_address(state_manager: &AtomaStateManager, user_id: i64, sui_address: String) -> Result<(), AtomaStateManagerError> {
+    ///    state_manager.store_sui_address(user_id, sui_address).await
+    /// }
+    /// ```
+    pub async fn update_sui_address(&self, user_id: i64, sui_address: String) -> Result<()> {
+        sqlx::query("UPDATE users SET sui_address = $1 WHERE id = $2")
```
jorgeantonio21, line 3618: Where are sui addresses inserted ?

jorgeantonio21 (commented):
LGTM

Please address the comments above

atoma-auth/src/auth.rs
Cifko: added user_id. It has to match the access_token user_id.

atoma-auth/src/auth.rs
```diff
@@ -355,6 +369,69 @@ impl Auth {
...
+                    .as_bytes()
+                    .to_vec(),
+            },
+        );
+
+        let intent_bcs = bcs::to_bytes(&intent_msg)?;
+        let message_hash = blake2b_hash(&intent_bcs);
```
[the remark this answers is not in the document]
Cifko: Blake2b and blake3 are not compatible. I have to do what typescript sdk is doing.
But we can probably change the blake2b that I use for password in this code. But that's for another PR.

atoma-state/src/state_manager.rs
Cifko: when the user registers

jorgeantonio21 (approved):
LGTM

PR #96, opened 2024-12-30
add traefik container to route requests to the proxy

### LINKED ISSUE #94: Add traefik container to route requests to the proxy at a specified url, with https
Ideally, we would have a single traefik container that is responsible to act as a reverse proxy both for the current Atoma proxy service and the Atoma dashboard, as well. 

By default, the proxy host domain should be https://api.atomacloud.com (which can be configurable).

fishonamos:
Closes #94

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,4 +1,32 @@
 services:
+  traefik:
+    image: traefik:v3.2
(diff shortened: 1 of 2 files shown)


jorgeantonio21:
Closed in favor of #106

PR #103, opened 2024-12-31
feat: get balance and stacks for user

Cifko:
Add endpoints to get stacks for the user, and the balance, both is for the dashboard billings page.


jorgeantonio21 (approved):
LGTM

PR #109, opened 2025-01-02
feat: return stack timestamp

Cifko:
Return stack timestamp alongside with the stack itself.
I don't want to add the timestamp to stack struct, because it would not reflect the event that's emitted.


jorgeantonio21 (approved):
LGTM

PR #126, opened 2025-01-08
feat: digest table and multiple user with the same wallet

Cifko:
- Enable multiple users to have the same wallet.
- Get rid of the timestamp in the balance table.
- Introduce new table with the usdc digests. Each digest can be processed only once.


jorgeantonio21 (approved):
LGTM

PR #128, opened 2025-01-08
refactor: nodes endpoints

francis2tm:


atoma-proxy/src/server/handlers/nodes.rs
```diff
@@ -0,0 +1,389 @@
...
+    responses(
+        (status = OK, description = "Node public address registered successfully", body = Value),
+        (status = INTERNAL_SERVER_ERROR, description = "Failed to register node public address")
+    )
+)]
+#[instrument(level = "info", skip_all)]
+pub async fn nodes_create(
+    State(state): State<ProxyState>,
+    headers: HeaderMap,
+    request: Request<Body>,
```
francis2tm, line 84: What's this Request Body? Is there a struct for this? -- Need it for docs

jorgeantonio21: The body should be a JSON that deserializes into the struct `NodePublicAddressAssignment`. We can definitely change it. Also we have a `signature` field in the headers, which we can add to the JSON body as well.

jorgeantonio21: The only problem with the signature in the JSON body is that the signature is then not the signature of the whole payload hash, but instead of its parts.

atoma-proxy/src/server/handlers/nodes.rs
```diff
@@ -0,0 +1,389 @@
...
+///
+/// This endpoint is specifically designed for confidential compute scenarios where
+/// requests need to be encrypted before being processed by nodes.
+///
+/// ## Errors
+///   - `INTERNAL_SERVER_ERROR` - Communication errors or missing node public keys
+///   - `SERVICE_UNAVAILABLE` - No nodes available for confidential compute
+#[utoipa::path(
+    get,
+    path = "/models/{model}",
```
jorgeantonio21, line 227: Should we add the `v1` ?

jorgeantonio21 (approved):
LGTM

Left a few comments

atoma-proxy/src/server/handlers/nodes.rs
jorgeantonio21: Not needed

jorgeantonio21 (approved):
LGTM

PR #130, opened 2025-01-09
chore: change anyhow to custom error on auth

Cifko:
This is for easier handling of the responses. E.g. user already registered now can throw 409 (conflict).

### DIFF
--- a/atoma-auth/src/auth.rs
+++ b/atoma-auth/src/auth.rs
@@ -1,7 +1,6 @@
 use std::{str::FromStr, sync::Arc};
 
-use anyhow::Result;
-use atoma_state::types::AtomaAtomaStateManagerEvent;
+use atoma_state::{types::AtomaAtomaStateManagerEvent, AtomaStateManagerError};
 use atoma_utils::hashing::blake2b_hash;
 use blake2::{
     digest::{consts::U32, generic_array::GenericArray},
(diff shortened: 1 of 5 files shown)


atoma-auth/src/auth.rs
```diff
@@ -44,6 +45,49 @@ pub struct Claims {
     refresh_token_hash: Option<String>,
 }
 
+#[derive(Error, Debug)]
+pub enum AuthError {
```
jorgeantonio21, line 49: Much better

atoma-proxy-service/src/handlers/auth.rs
```diff
@@ -268,7 +269,10 @@ pub(crate) async fn register(
         .await
         .map_err(|e| {
             error!("Failed to register user: {:?}", e);
-            StatusCode::INTERNAL_SERVER_ERROR
+            match e {
+                AuthError::UserAlreadyRegistered => StatusCode::CONFLICT,
+                _ => StatusCode::INTERNAL_SERVER_ERROR,
```
jorgeantonio21, line 274: It would be great to have a proper error messaging, maybe for another PR

jorgeantonio21 (approved):
LGTM

atoma-proxy-service/src/handlers/auth.rs
Cifko: yeah, it would be nice if we return the proper statuscodes, and not internal server error most of the times
IMHO it should be another PR

PR #131, opened 2025-01-09
chore: change default stacks size to buy to 1M

Cifko:
Buy 1M tokens per stacks.


Cifko:
Can't be done just on proxy.

PR #132, opened 2025-01-09
chore: change default stacks size to buy to 1M

Cifko:
Change the stacks size to buy to 1 million compute units.


jorgeantonio21 (approved):
LGTM

PR #134, opened 2025-01-09
fix: balance change can be zero

Cifko:
The balance should be able to drop to zero. It was required to be non-zero.


jorgeantonio21 (approved):
LGTM

PR #137, opened 2025-01-09
feat: return proper login error

Cifko:
Return 401 if the user doesn't exist or the passwords doesn't match.


jorgeantonio21 (approved):
LGTM

PR #140, opened 2025-01-10
feat: refactor to update tokens correctly on middleware failure

jorgeantonio21:


atoma-proxy/src/server/middleware.rs
```diff
@@ -255,64 +249,70 @@ pub async fn authenticate_middleware(
...
+            &body_json,
+            &endpoint,
+        )
+        .await?;
+
+    // Selects an appropriate node to process the request (if there is no available node for the stacks the proxy holds, it buys a new stack)
+    //
+    // NOTE: IF `optional_stack` is Some, this means that the proxy has locked enough compute units for the request, within the state manager, already.
+    // In this case, this method cannot error (as it just returns the underlying stack data). Otherwise, it will try to buy a new stack.
+    // If this method succeeds, this means that the proxy has locked enough compute units for the request, within the state manager, already.
```
Cifko, line 270:
```suggestion
    // If this method succeeds, this means that the proxy has locked enough compute units for the request, within the state manager.
```

atoma-proxy/src/server/middleware.rs
```diff
@@ -824,124 +993,158 @@ pub(crate) mod auth {
...
-            }
-            _ => Err(AtomaProxyError::NotFound {
-                message: format!("Invalid or unsupported endpoint: {}", endpoint),
-                endpoint: endpoint.to_string(),
-            }),
+            req_parts
+                .headers
+                .insert(constants::TX_DIGEST, tx_digest_header);
         }
+        let request_model = body_json.get("model").and_then(|m| m.as_str()).ok_or(
```
Cifko, line 1129: move "model" into constant (if we don't have one already)

atoma-proxy/src/server/middleware.rs
```diff
@@ -824,124 +993,158 @@ pub(crate) mod auth {
...
+            AtomaProxyError::InvalidBody {
+                message: "Model not found".to_string(),
```
Cifko, line 1131: you can reuse the above constant over here as well

Cifko (approved):

atoma-proxy/src/server/middleware.rs
jorgeantonio21: Oh nice catch

atoma-proxy/src/server/middleware.rs
jorgeantonio21: Addressed

PR #145, opened 2025-01-13
refactor: add nodes create lock endpoint

francis2tm:
- Introduced a new endpoint for creating a node lock for confidential compute.
- Updated the OpenAPI documentation to reflect the new endpoint and request/response structures.
- Refactored existing code to replace the previous models retrieve endpoint with the new create lock functionality.


jorgeantonio21 (approved):
LGTM

PR #150, opened 2025-01-16
fix: users without balance should fail when deducting

Cifko:
When the user was not present the query didn't fail, it just updated zero rows. Now we check if the row was affected.


jorgeantonio21 (approved):
This makes sense,

LGTM

PR #151, opened 2025-01-17
feat: google oauth and zklogin

Cifko:
Add google oauth
Salt storage for zklogin
Each usdc topup from zklogin have to provide a zklogin proof signing the transaction digest. The proof needs to be from the same address as the topup (same requirements we have for the normal sui address). This way we don't need to store the zkaddress.


atoma-auth/src/auth.rs
```diff
@@ -496,6 +542,72 @@ impl Auth {
         Ok(())
     }
 
+    async fn get_zk_address(zk_login_signature: &str, tx_digest: &str) -> Result<String> {
```
jorgeantonio21, line 545: Add docs and instrumentation

atoma-auth/src/auth.rs
```diff
@@ -496,6 +542,72 @@ impl Auth {
...
+        let user_signature = UserSignature::from_base64(zk_login_signature)?;
+        if let UserSignature::ZkLogin(zk_login_authenticator) = user_signature {
+            // The message is the transaction digest that we claim is ours
```
jorgeantonio21, line 548:
```suggestion
            // The message is the transaction digest that the user claims its theirs
```

atoma-auth/src/auth.rs
```diff
@@ -496,6 +542,72 @@ impl Auth {
...
+            let intent_msg = IntentMessage::new(
+                Intent::personal_message(),
```
jorgeantonio21, line 550: I am confused by this

atoma-auth/src/auth.rs
```diff
@@ -496,6 +542,72 @@ impl Auth {
...
+                    ))
+                })?;
+            let public_id =
+                ZkLoginPublicIdentifier::new(google::ISS, &address_seed).map_err(|e| {
+                    AuthError::FailedToParseSignature(format!(
+                        "Failed to parse public identifier: {}",
+                        e
+                    ))
+                })?;
+            let public_key = PublicKey::ZkLogin(public_id);
```
jorgeantonio21, line 601: Shouldn't this public key be the same as the one that verifies the signature, above ?

atoma-auth/src/auth.rs
```diff
@@ -496,6 +542,72 @@ impl Auth {
...
+            let sui_address = SuiAddress::from(&public_key);
```
jorgeantonio21, line 602: I wonder if we couldn't just construct the `sui_address` from the public key that is used to verify the given signature above

atoma-auth/src/auth.rs
```diff
@@ -719,12 +848,15 @@ active_address: "0x939cfcc7fcbc71ce983203bcb36fa498901932ab9293dfa2b271203e71603
...
         let (state_manager_sender, state_manager_receiver) = flume::unbounded();
 
         let sui_config = AtomaSuiConfig::from_file_path(get_config_path());
         let sui = crate::Sui::new(&sui_config).await.unwrap();
-        let auth = Auth::new(config, state_manager_sender, Arc::new(RwLock::new(sui)));
+        let auth = Auth::new(config, state_manager_sender, Arc::new(RwLock::new(sui)))
+            .await
+            .unwrap();
         (auth, state_manager_receiver)
     }
```
jorgeantonio21, line 862: No tests for the current zkLogin logic ?

atoma-auth/src/auth.rs
```diff
@@ -571,18 +688,30 @@ impl Auth {
...
-                .send(AtomaAtomaStateManagerEvent::ConfirmUser {
-                    sui_address: sender.to_string(),
-                    user_id: claims.user_id,
-                    result_sender,
-                })?;
-            let is_their_wallet = result_receiver.await??;
-
-            if !is_their_wallet {
-                return Err(AuthError::PaymentNotForThisUser);
+            match proof_signature {
```
jorgeantonio21, line 691: Can you explain where does how is this `proof_signature` is generated ? And in which scenarios is it present in the current method body scope and when it is not ?

atoma-auth/src/google.rs
```diff
@@ -0,0 +1,134 @@
...
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use thiserror::Error;
+
+pub const ISS: &str = "https://accounts.google.com";
+
+/// Claims struct for Google ID tokens
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Claims {
+    iss: String,               // Issuer
```
jorgeantonio21, line 16: Can we add proper documentation and not comments ?

atoma-auth/src/google.rs
```diff
@@ -0,0 +1,134 @@
...
+    #[error("Public keys malformed")]
+    MalformedPublicKeys,
+    #[error("Email not found in token claims")]
+    EmailNotFound,
+}
+
+type Result<T> = std::result::Result<T, GoogleError>;
+
+/// Convert a JWK to a DecodingKey
+fn jwk_to_decoding_key(n: &str, e: &str) -> Result<DecodingKey> {
```
jorgeantonio21, line 48: Maybe it is worth adding instrumentation to this ?

atoma-auth/src/google.rs
```diff
@@ -0,0 +1,134 @@
...
+    let pem = public_key.to_pkcs1_pem(rsa::pkcs1::LineEnding::LF)?;
+
+    // Create the DecodingKey
+    let decoding_key = DecodingKey::from_rsa_pem(pem.as_bytes())?;
+
+    Ok(decoding_key)
+}
+
+/// Fetch Google's public keys from the JWKS endpoint
+pub async fn fetch_google_public_keys() -> Result<HashMap<String, DecodingKey>> {
```
jorgeantonio21, line 70: Add instrumentation ?

atoma-auth/src/google.rs
```diff
@@ -0,0 +1,134 @@
...
+    let jwks_url = "https://www.googleapis.com/oauth2/v3/certs";
```
jorgeantonio21, line 71: Add this a public constant

atoma-auth/src/google.rs
```diff
@@ -0,0 +1,134 @@
...
+    let client = Client::new();
+    let res = client.get(jwks_url).send().await?;
+    let jwks: Value = res.json().await?;
+    let keys = jwks["keys"]
```
jorgeantonio21, line 75: Add a constant for this `KEYS`. Also, I suggest using `.get(KEYS)` to avoid panics on the proxy

atoma-auth/src/google.rs
```diff
@@ -0,0 +1,134 @@
...
+        .as_array()
+        .ok_or(GoogleError::MalformedPublicKeys)?;
+
+    let mut public_keys = HashMap::new();
+    for key in keys {
+        let kid = key["kid"]
```
jorgeantonio21, line 81: See previous comment

atoma-auth/src/google.rs
```diff
@@ -0,0 +1,134 @@
...
+            .as_str()
+            .ok_or(GoogleError::MalformedPublicKeys)?
+            .to_string();
+        let n = key["n"]
```
jorgeantonio21, line 85: See comments above

atoma-auth/src/google.rs
```diff
@@ -0,0 +1,134 @@
...
+            .as_str()
+            .ok_or(GoogleError::MalformedPublicKeys)?
+            .to_string();
+        let e = key["e"]
```
jorgeantonio21, line 89: See comments above

atoma-auth/src/google.rs
```diff
@@ -0,0 +1,134 @@
...
+///
+/// # Arguments
+///
+/// * `id_token` - The ID token to verify
+/// * `audience` - The expected audience of the token
+///
+/// # Returns
+///
+/// A `Result` containing the verified claims if successful, or an error if verification failed
+pub fn verify_google_id_token(
```
jorgeantonio21, line 109: Shall we add instrumentation here ?

atoma-auth/src/auth.rs
Cifko: No. Private key for zklogin transaction doesn't correspond to the address. The address is generated from the zk data.

atoma-auth/src/auth.rs
Cifko: Refactored into new function.
The original TS code where this message is created https://github.com/MystenLabs/ts-sdks/blob/main/packages/typescript/src/cryptography/intent.ts

atoma-auth/src/auth.rs
jorgeantonio21: I see

jorgeantonio21 (approved):
LGTM

PR #160, opened 2025-01-23
feat: add zklogin support from sui wallet

Cifko:
Add support for wallet that are zklogin in sui wallet.
Unified signature check for both simple and zklogin signatures.
To get the iss it uses internal function of sui-rust-sdk, so we have to be aware that this can break in the future. But it's currently the best way to get the iss.


jorgeantonio21 (approved):
LGTM

PR #161, opened 2025-01-23
fix: missing network in docker for prover

Cifko:
Frontend could not connect to the backend, because they ended up on different network.


jorgeantonio21 (approved):
LGTM

PR #172, opened 2025-01-29
feat: update max tokens according to openai api specs to max_completion_tokens

jorgeantonio21:


atoma-proxy/docs/openapi.yml
```diff
@@ -568,12 +568,20 @@ components:
...
+          description: The maximum number of tokens to generate in the chat completion
+          example: 2048
         max_tokens:
           type:
           - integer
           - 'null'
           format: int32
           description: The maximum number of tokens to generate in the chat completion
+          deprecated: true
           example: 2048
```
maschad, line 585: Very minor but do you think this should be 4096 as well?

atoma-proxy/src/server/mod.rs
```diff
@@ -21,6 +21,9 @@ pub type Result<T> = std::result::Result<T, error::AtomaProxyError>;
 /// The max_tokens field in the request payload.
 pub(crate) const MAX_TOKENS: &str = "max_tokens";
```
maschad, line 22: Should we add a deprecation notice here as well?

maschad (approved):
LGTM but I left two small comments,

Also I think we should have follow up issues both here and on https://github.com/atoma-network/atoma-node

To remove the deprecated code at some point

atoma-proxy/src/server/mod.rs
jorgeantonio21: Nice observation, added

atoma-proxy/docs/openapi.yml
jorgeantonio21: I believe I addressed this comment, please confirm on the most recent changes

maschad: Nice! resolving

PR #177, opened 2025-01-30
feat: handle partial keep alive chunks

### LINKED ISSUE #178: Address partial streamed chat completions keep alive chunks

jorgeantonio21:
Fixes #178

### DIFF
--- a/atoma-proxy/src/server/streamer.rs
+++ b/atoma-proxy/src/server/streamer.rs
@@ -11,7 +11,7 @@ use std::{
     task::{Context, Poll},
     time::Instant,
 };
-use tracing::{error, info, instrument};
+use tracing::{error, info, instrument, warn};
 
 use crate::server::handlers::{chat_completions::CHAT_COMPLETIONS_PATH, update_state_manager};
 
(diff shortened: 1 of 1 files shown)


atoma-proxy/src/server/streamer.rs
```diff
@@ -298,6 +304,27 @@ impl Stream for Streamer {
...
+                        self.keep_alive_pos += chunk_str.len();
+
+                        if self.keep_alive_pos == KEEP_ALIVE_CHUNK_STR.len() {
+                            // Full keep-alive received
+                            self.keep_alive_pos = 0;
+                            return Poll::Pending;
+                        }
+                        return Poll::Pending;
+                    } else if self.keep_alive_pos > 0 {
+                        // Reset position if we had partial match but current chunk doesn't continue it
```
Cifko, line 318: this should never happen, the chunk (and their pieces) are ordered, so I think we should return error when this happens

jorgeantonio21: Nice observation, addressed

Cifko (approved):

PR #189, opened 2025-01-31
feat: update cargo lock file and resolve some unresolved clippy issues

jorgeantonio21:


francis2tm (approved):

maschad:
@jorgeantonio21  why is this necessary? It seems like the previous builds didn't have any `clippy` related issues

atoma-auth/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![allow(clippy::doc_markdown)]
 #![allow(clippy::cast_sign_loss)]
 #![allow(clippy::cast_possible_wrap)]
+#![allow(clippy::module_name_repetitions)]
```
maschad, line 4: we shouldn't need to do this if your local rust is `1.84.0` which is the current LTS

maschad (changes requested):
I approve the updated `Cargo.lock` but we shouldn't need the clippy allowances

atoma-auth/src/lib.rs
jorgeantonio21: Removed all of instances of this

maschad (approved):
LGTM

PR #195, opened 2025-02-03
feat: add multiple TEE modules for handling node key rotation events for different CPU ISAs

bredamatt:


atoma-state/Cargo.toml
```diff
@@ -5,11 +5,13 @@ edition.workspace = true
 license.workspace = true
 
 [dependencies]
+atoma-confidential = { workspace = true, features = ["sev-snp", "tdx"] }
```
bredamatt, line 8: had to include them both here - and this is what throws compilation issues when doing `cargo c` etc. on a macb, because the sev-snp crate is feature-gated for Linux.

Cargo.toml
```diff
@@ -11,6 +11,7 @@ license = "Apache-2.0"
 anyhow = "1.0.91"
 async-trait = "0.1.86"
 atoma-auth = { path = "./atoma-auth" }
+atoma-confidential = { git = "https://github.com/atoma-network/atoma-node.git", package = "atoma-confidential", branch = "main", features = ["sev-snp", "tdx"] }
```
jorgeantonio21, line 14: Can we use a tool to format the toml files ?

jorgeantonio21: This means that only proxies running on Linux (which is probably fine if it is deployed through docker) but on `x86_64`architecture can run this code, right ?

jorgeantonio21 (commented):
It is looking great, if I understand correctly this code only compiles on x86_64 archs ?

Cargo.toml
bredamatt: Regarding `.toml` format, we can do something like this:
```
[dependencies.atoma-confidential]
git = "url_here"
package = "package_here"
...
```
which is in sync with the conventions defined here: https://doc.rust-lang.org/style-guide/cargo.html.

Regarding the architecture, I think that only `x86_64` supports AMD SNP and Intel CPUs, so yes, correct. If ARM TrustZone is added down the line, then this would require a separate target, i.e. `arm64`.

Furthermore, regarding Linux, yes if the proxy runs in a container which uses a Linux base image which supports the capabilities required by AMD SEV SNP and Intel TDX, then it will work.

bredamatt: Also, note that the above implies you are compiling with the `confidential` feature. It will compile without this feature on any machine for now, but naturally, you won't have the ability to verify attestation reports of confidential compute environments without that feature enabled.

PR #201, opened 2025-02-03
feat: telemetry

francis2tm:


atoma-proxy/src/server/middleware.rs
```diff
@@ -457,6 +460,41 @@ pub async fn confidential_compute_middleware(
     Ok(next.run(req).await)
 }
 
+/// Middleware to track request latency and middleware timing
+#[instrument(skip_all)]
+pub async fn track_metrics(request: Request<Body>, next: Next) -> Result<Response> {
```
jorgeantonio21, line 465: Nice

atoma-proxy/src/server/middleware.rs
```diff
@@ -457,6 +460,41 @@ pub async fn confidential_compute_middleware(
...
+
+    // Record the metrics
+    record_request_duration(request_duration);
+
+    Ok(response)
+}
+
+/// Middleware to track database operations timing
+#[instrument(skip_all)]
+pub async fn track_db_metrics(
```
jorgeantonio21, line 480: Can't this middleware layer method and the previous one be combined into one ?

atoma-proxy/src/server/middleware.rs
```diff
@@ -457,6 +460,41 @@ pub async fn confidential_compute_middleware(
     Ok(next.run(req).await)
 }
 
+/// Middleware to track request latency and middleware timing
+#[instrument(skip_all)]
+pub async fn track_metrics(request: Request<Body>, next: Next) -> Result<Response> {
+    let start = Instant::now();
```
jorgeantonio21, line 466: Duration time should be split across successful and failed requests. I suggest tracking as well the number of failed vs successful requests as part of the metrics pipeline

atoma-proxy/src/telemetry.rs
```diff
@@ -0,0 +1,195 @@
...
+use tracing_opentelemetry::OpenTelemetryLayer;
+use tracing_subscriber::{
+    fmt::{self, format::FmtSpan, time::UtcTime},
+    layer::SubscriberExt,
+    util::SubscriberInitExt,
+    EnvFilter, Registry,
+};
+
+const LOG_FILE: &str = "atoma-proxy-service.log";
+// Default Grafana OTLP endpoint if not specified in environment
```
jorgeantonio21, line 29: Add empty line

atoma-proxy/src/telemetry.rs
```diff
@@ -0,0 +1,195 @@
...
+// Define metric instruments
+static REQUEST_DURATION: Lazy<Histogram<f64>> = Lazy::new(|| {
+    GLOBAL_METER
+        .f64_histogram("request_duration")
+        .with_description("Total request latency in seconds")
+        .with_unit(Unit::new("s"))
+        .init()
+});
+
+static MIDDLEWARE_DURATION: Lazy<Histogram<f64>> = Lazy::new(|| {
```
jorgeantonio21, line 47: Add some docs

atoma-proxy/src/telemetry.rs
```diff
@@ -0,0 +1,195 @@
...
+    GLOBAL_METER
+        .f64_histogram("middleware_duration")
+        .with_description("Middleware execution time in seconds")
+        .with_unit(Unit::new("s"))
+        .init()
+});
+
+static DB_READ_DURATION: Lazy<Histogram<f64>> = Lazy::new(|| {
```
jorgeantonio21, line 55: Where are we using this ?

jorgeantonio21: Add docs

atoma-proxy/src/telemetry.rs
```diff
@@ -0,0 +1,195 @@
...
+    GLOBAL_METER
+        .f64_histogram("db_read_duration")
+        .with_description("Database read operation duration in seconds")
+        .with_unit(Unit::new("s"))
+        .init()
+});
+
+static DB_WRITE_DURATION: Lazy<Histogram<f64>> = Lazy::new(|| {
```
jorgeantonio21, line 63: Where are we using this ? 

Add docs

atoma-proxy/src/telemetry.rs
```diff
@@ -0,0 +1,195 @@
...
+    GLOBAL_METER
+        .f64_histogram("db_write_duration")
+        .with_description("Database write operation duration in seconds")
+        .with_unit(Unit::new("s"))
+        .init()
+});
+
+static DB_READ_COUNT: Lazy<Counter<u64>> = Lazy::new(|| {
```
jorgeantonio21, line 71: Add docs

atoma-proxy/src/telemetry.rs
```diff
@@ -0,0 +1,195 @@
...
+    GLOBAL_METER
+        .u64_counter("db_read_count")
+        .with_description("Number of database read operations")
+        .init()
+});
+
+static DB_WRITE_COUNT: Lazy<Counter<u64>> = Lazy::new(|| {
```
jorgeantonio21, line 78: Add docs

atoma-proxy/src/telemetry.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// Initialize metrics with OpenTelemetry SDK
+fn init_metrics() -> sdkmetrics::MeterProvider {
+    sdkmetrics::MeterProvider::builder()
+        .with_resource(RESOURCE.clone())
+        .build()
+}
+
+/// Initialize tracing with OpenTelemetry SDK
+fn init_traces() -> Result<sdktrace::Tracer> {
```
jorgeantonio21, line 93: Maybe good to have proper instrumentation and logging

atoma-proxy/src/server/middleware.rs
jorgeantonio21: Also from a statistically point of view, this seems to be incoherent, as we are storing all these metrics as a part of a single distribution that is common to all endpoints, when this is not the case. Chat completions vs node creation vs image gen have different distributions, and therefore we should collect metrics per available `handler`. Unless, I am actually wrong in thinking this, and open-telemetry already decouples the metric collection for different handler endpoints (which I would be surprised as it does not so happen with the `prometheus` library)

jorgeantonio21 (commented):
Looking very good, 

Left some comments though

Cargo.lock
```diff
@@ -1889,7 +1889,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1145d32e826a7748b69ee8fc62d3e6355ff7f1051df53141e7048162fc90481b"
 dependencies = [
  "data-encoding",
- "syn 2.0.96",
+ "syn 1.0.109",
```
jorgeantonio21, line 1892: It's dumping versions ?

maschad (changes requested):
Great work on this @francis2tm  🚀 

I agree with @jorgeantonio21 suggestions but also I think we should consider upgrading to the latest `opentelemetry` sdks as many of these methods are being deprecated, and it's easier to to iron out these breaking changes sooner rather than later.

[Dependabot CI](https://github.com/atoma-network/atoma-proxy/pull/196) is already complaining when trying to bump them

atoma-proxy/src/telemetry.rs
maschad: What do you mean by this @jorgeantonio21 ?

PR #203, opened 2025-02-06
feat: add cors layer to proxy

Cifko:
Add cors layer to proxy, so it can be fetched from frontends.
Remove requirements for bearer token from `/v1/models` endpoint.


jorgeantonio21 (approved):
LGTM

PR #206, opened 2025-02-06
build: update grafana config

maschad:


docker-compose.yaml
```diff
@@ -180,9 +180,7 @@ services:
       - "3000:3000"
     volumes:
       - grafana_data:/var/lib/grafana
-    environment:
-      - GF_AUTH_ANONYMOUS_ENABLED=false
-      - GF_SECURITY_ADMIN_PASSWORD=admin
+    env_file: .env
```
jorgeantonio21, line 183: Add the required missing values to the `.env.example`

jorgeantonio21 (commented):
LGTM

Please add necessary variables to `.env.example`

docker-compose.yaml
maschad: Good catch! done in 6ac73ae

jorgeantonio21 (approved):
LGTM

PR #219, opened 2025-02-11
feat: update openai api to handle contents in array form

jorgeantonio21:


atoma-proxy/docs/openapi.yml
```diff
@@ -472,6 +477,11 @@ components:
           example: meta-llama/Llama-3.3-70B-Instruct
     ChatCompletionChunkChoice:
       type: object
+      description: |-
+        Represents the chat completion chunk choice.
+
+        This is used to represent the chat completion chunk choice in the chat completion request.
+        It can be either a chat completion chunk delta or a chat completion chunk choice.
```
maschad, line 484: Should we have the properties documented here? the `index` and `delta` are required so at the very least I think they should be

atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -795,33 +808,70 @@ impl RequestModel for RequestModelChatCompletions {
...
+                                total_num_tokens += num_tokens;
+                                // add 2 tokens as a safety margin, for start and end message delimiters
+                                total_num_tokens += 2;
+                                // add 1 token as a safety margin, for the role name of the message
+                                total_num_tokens += 1;
+                                // add the max completion tokens, to account for the response
+                                total_num_tokens += self.max_completion_tokens;
+                            }
+                            MessageContentPart::Image { .. } => {
+                                // TODO: Ensure that for image content parts, we have a way to estimate the number of tokens,
```
maschad, line 858: We should create an issue for this

atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -795,33 +808,70 @@ impl RequestModel for RequestModelChatCompletions {
...
+                    total_num_tokens += 2;
+                    // add 1 token as a safety margin, for the role name of the message
+                    total_num_tokens += 1;
+                    // add the max completion tokens, to account for the response
+                    total_num_tokens += self.max_completion_tokens;
+                }
+                MessageContent::Array(parts) => {
+                    for part in parts {
+                        match part {
+                            MessageContentPart::Text { text, .. } => {
```
maschad, line 840: if you have a conversation with 3 messages, and `max_completion_tokens` is 1000, you're adding:

Message 1: +1000 tokens
Message 2: +1000 tokens
Message 3: +1000 tokens

So you end up estimating 3000 completion tokens, is that correct?

atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -795,33 +808,70 @@ impl RequestModel for RequestModelChatCompletions {
...
-                })?
-                .get_ids()
-                .len() as u64;
-
-            total_num_tokens += num_tokens;
-            // add 2 tokens as a safety margin, for start and end message delimiters
-            total_num_tokens += 2;
-            // add 1 token as a safety margin, for the role name of the message
-            total_num_tokens += 1;
+            match content {
```
maschad, line 818: There's some repetition here around `total_num_tokens`, perhaps it could be refactored to something like:

```rust

    // Helper function to count tokens for a text string
    let count_text_tokens = |text: &str| -> Result<u64> {
        Ok(tokenizer
            .encode(text, true)
            .map_err(|err| AtomaProxyError::InternalError {
                message: format!("Failed to encode message: {err:?}"),
                endpoint: CHAT_COMPLETIONS_PATH.to_string(),
            })?
            .get_ids()
            .len() as u64)
    };

    let mut prompt_tokens = 0;

    // Process each message
    for message in &self.messages {
        let content = message
            .get(CONTENT_KEY)
            .and_then(|content| MessageContent::deserialize(content).ok())
            .ok_or_else(|| AtomaProxyError::RequestError {
                message: "Missing or invalid message content".to_string(),
                endpoint: CHAT_COMPLETIONS_PATH.to_string(),
            })?;

        // Add fixed token overhead per message
        prompt_tokens += 3; // 2 for delimiters + 1 for role

        // Count tokens in content
        match content {
            MessageContent::Text(text) => {
                prompt_tokens += count_text_tokens(&text)?;
            }
            MessageContent::Array(parts) => {
                for part in parts {
                    match part {
                        MessageContentPart::Text { text, .. } => {
                            prompt_tokens += count_text_tokens(&text)?;
                        }
                        MessageContentPart::Image { .. } => {
                            // Add a conservative estimate for image tokens
                            // This should be replaced with actual image token calculation
                            prompt_tokens += 100; // Example placeholder
                        }
                    }
                }
            }
        }
    }

    // Add completion tokens only once
    Ok(prompt_tokens + self.max_completion_tokens)
    ```

File: atoma-proxy/src/server/handlers/chat_completions.rs [hunk 5]
```diff
@@ -935,22 +989,244 @@ pub struct ChatCompletionRequest {
...
+            }
+        }
+    }
+}
+
+/// Represents the image URL of a message content part.
+///
+/// This is used to represent the image URL of a message content part in the chat completion request.
+/// It can be either a URL or a base64 encoded image data.
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
```
maschad, line 1143: might benefit from deriving `Clone` if they need to be duplicated in memory.

atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -935,22 +989,244 @@ pub struct ChatCompletionRequest {
...
+/// This is used to represent the function that the model called in the chat completion request.
+/// It can be either a function or a tool call.
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
+pub struct ToolCallFunction {
+    /// The name of the function to call.
+    name: String,
+    /// The arguments to call the function with, as generated by the model in JSON format.
+    /// Note that the model does not always generate valid JSON, and may hallucinate parameters not defined by your function schema.
+    /// Validate the arguments in your code before calling your function.
+    arguments: Value,
```
maschad, line 1173: We could use a stronger type for this in the future such as:

```rust
// Represents the structure of function arguments
#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
#[serde(untagged)]
pub enum FunctionArguments {
    /// Arguments in a structured format
    Structured(HashMap<String, JsonValue>),
    
    /// Raw JSON string for cases where the model outputs malformed JSON
    RawString(String),
}

/// Represents the function that the model called.
#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
pub struct ToolCallFunction {
    /// The name of the function to call.
    pub name: String,
    
    /// The arguments to call the function with, as generated by the model.
    /// Note: The model may generate invalid JSON or hallucinate parameters.
    /// Always validate arguments before use.
    #[serde(deserialize_with = "deserialize_function_arguments")]
    pub arguments: FunctionArguments,
}
```

What do you think?

maschad (changes requested):
Looking good but left some comments

atoma-proxy/src/server/handlers/chat_completions.rs
jorgeantonio21: This makes a lot of sense, that said, after consulting the official OpenAI API docs it seems that they consider this to be of type `string`, see ref:

https://platform.openai.com/docs/api-reference/chat/object

atoma-proxy/docs/openapi.yml
jorgeantonio21: I removed this comment, since we are already documenting the chunk delta, etc

atoma-proxy/src/server/handlers/chat_completions.rs
jorgeantonio21: Good idea, created #223

atoma-proxy/src/server/handlers/chat_completions.rs
jorgeantonio21: I love this suggestion. I address it, I had to change some of the code, since you are not accounting for the fact that `self.max_completion_tokens` and the overhead should be multiplied by the number of messages in the `array` case.

maschad: Looks good, can we just write a test for this function since it's critical functionality?

maschad (commented):
Looks good, just [one comment](https://github.com/atoma-network/atoma-proxy/pull/219/files#r1952824518) left and then we can merge

atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -791,283 +828,962 @@ impl RequestModel for RequestModelChatCompletions {
...
+        pub enum MessageContentPart {
+            #[serde(rename(serialize = "text", deserialize = "text"))]
+            Text {
+                /// The type of the content part.
+                #[serde(rename(serialize = "type", deserialize = "type"))]
+                r#type: String,
+                /// The text content.
+                text: String,
+            },
+            #[serde(rename(serialize = "image", deserialize = "image"))]
```
Cifko, line 1361: can we add the audio as well?

atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -791,283 +828,962 @@ impl RequestModel for RequestModelChatCompletions {
...
+        };
+
         let mut total_num_tokens = 0;
 
         for message in &self.messages {
             let content = message
-                .get("content")
-                .and_then(|content| content.as_str())
+                .get(CONTENT_KEY)
+                .and_then(|content| MessageContent::deserialize(content).ok())
```
Cifko, line 848: this will fail for valid openai request, where assistent message has an array with refusal only (no text)
from the specs `An array of content parts with a defined type. Can be one or more of type text, or exactly one of type refusal.` I'm not sure about the other messages because the openai lacks the specs for the content arrays on messages tool, system, developer
or we can add the support later

atoma-proxy/src/server/handlers/chat_completions.rs
jorgeantonio21: Added a test suite

atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -791,283 +828,962 @@ impl RequestModel for RequestModelChatCompletions {
...
+                                // TODO: Ensure that for image content parts, we have a way to estimate the number of tokens,
+                                // which can depend on the size of the image and the output description.
+                                continue;
+                            }
+                        }
+                    }
+                }
+            }
+            // add the max completion tokens, to account for the response
+            total_num_tokens += num_messages * self.max_completion_tokens;
```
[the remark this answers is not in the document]
jorgeantonio21: This was being incorrectly calculated, I have a made the final summation at the end of the function, seems pretty readable to me

atoma-proxy/src/server/handlers/chat_completions.rs
jorgeantonio21: Not in this PR, we will need to add support later

atoma-proxy/src/server/handlers/chat_completions.rs
jorgeantonio21: We currently don't support refusal only (as vLLM itself does not support it), so we might as well error on here

atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -791,283 +828,962 @@ impl RequestModel for RequestModelChatCompletions {
...
-                .get_ids()
-                .len() as u64;
-
-            total_num_tokens += num_tokens;
-            // add 2 tokens as a safety margin, for start and end message delimiters
-            total_num_tokens += 2;
-            // add 1 token as a safety margin, for the role name of the message
-            total_num_tokens += 1;
+            // Add fixed token overhead per message
+            total_num_tokens += 3; // 2 for delimiters + 1 for role
```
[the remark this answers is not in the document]
jorgeantonio21: I changed it appropriately, also added tests

atoma-proxy/docs/openapi.yml
```diff
@@ -422,8 +422,22 @@ paths:
       - bearerAuth: []
 components:
   schemas:
+    Audio:
```
maschad, line 425: If [we currently don't support audio](https://github.com/atoma-network/atoma-proxy/pull/219#discussion_r1953112331) then we should only document it when it's available

maschad (commented):
LGTM but this one comment

atoma-proxy/docs/openapi.yml
jorgeantonio21: True, removed

maschad (approved):
LGTM

maschad (approved):

PR #222, opened 2025-02-11
feat: node scores computation and storage

jorgeantonio21:


atoma-state/src/handlers.rs
```diff
@@ -1767,3 +1908,1027 @@ mod utils {
...
+    ///
+    /// # Notes
+    ///
+    /// - The function uses the GPU with the lowest free memory percentage to calculate the score,
+    ///   as this represents the most constrained resource
+    /// - The calculation assumes that higher VRAM utilization is better for performance
+    /// - Both input slices must be of equal length, representing the same GPUs in the same order
+    /// - Returns -1.0 if the input slices are empty (TODO: We will introduce other accelerators metrics in the future)
+    #[allow(clippy::cast_precision_loss)]
+    fn compute_gpu_vram_utilization_score(
```
maschad, line 2151: Given we have [some precedents](https://github.com/maschad/atoma-proxy/blob/main/atoma-state/src/state_manager.rs#L3549) for how much memory a model might use, do you think we could pass in another param here such as `model_required_memory: i64`  and then calculate score based on how many model instances could fit ? I know that the `TODO` below we should be moving towards KV cache considerations so ultimately this could be model agnostic, but interested to hear your thoughts.

I am envision something like

```rust
 if gpu_memory_free.is_empty() || gpu_memory_total.is_empty() {
        return -1.0;
    }
    
    // Calculate score based on how many model instances could fit
    gpu_vram_weight * gpu_memory_free
        .iter()
        .map(|&free| (free as f64) / (model_required_memory as f64))
        .sum::<f64>() / (gpu_memory_free.len() as f64)
 ```

File: atoma-state/src/handlers.rs [hunk 2]
```diff
@@ -1767,3 +1908,1027 @@ mod utils {
...
+    /// // is 75% of the way between threshold (80.0) and max (100.0)
+    /// ```
+    ///
+    /// # Notes
+    ///
+    /// - Returns 0.0 if the input slice is empty
+    /// - Uses partial_cmp for floating point comparisons, defaulting to 0.0 if comparison fails
+    /// - The linear scaling provides a smooth transition between optimal and critical temperatures
+    /// - Returns -1.0 if the input slice is empty (TODO: We will introduce other accelerators metrics in the future)
+    fn compute_gpu_temp_score(
```
maschad, line 2269: I think we could be more flexible here, the current implementation uses fixed thresholds. We could make this more flexible by considering GPU-specific characteristics:


```rust
struct GpuSpecs {
    max_temp: f64,
    tdp: f64,
    optimal_temp: f64,
}

fn compute_gpu_temp_score(
    gpu_temperatures: &[f64],
    gpu_specs: &[GpuSpecs],
    gpu_temp_weight: f64,
) -> f64 {
    if gpu_temperatures.is_empty() {
        return -1.0;
    }
    
    gpu_temp_weight * gpu_temperatures
        .iter()
        .zip(gpu_specs)
        .map(|(&temp, spec)| {
            if temp <= spec.optimal_temp {
                1.0
            } else if temp >= spec.max_temp {
                0.0
            } else {
                1.0 - ((temp - spec.optimal_temp) / (spec.max_temp - spec.optimal_temp))
            }
        })
        .min_by(|a, b| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal))
        .unwrap_or(0.0)
}
  ```

File: atoma-state/src/handlers.rs [hunk 3]
```diff
@@ -1767,3 +1908,1027 @@ mod utils {
...
+    /// // 2. Final score = 0.333 * 0.2 = 0.066
+    /// ```
+    ///
+    /// # Notes
+    ///
+    /// - Uses an inverse relationship to ensure higher traffic results in lower scores
+    /// - The denominator adds 1 to prevent division by zero
+    /// - The score is part of a larger node performance evaluation system
+    #[allow(clippy::cast_precision_loss)]
+    fn compute_network_score(network_rx: i64, network_tx: i64, network_score_weight: f64) -> f64 {
```
maschad, line 2546: I think we also need to consider the bandwidth of the node as well, higher traffic on wider bandwidths may still have greater capacity for accepting work. e.g.

```rust
fn compute_network_score(
    network_rx: i64,
    network_tx: i64,
    network_capacity: i64,  // New parameter
    network_score_weight: f64
) -> f64 {
    let total_usage = network_rx as f64 + network_tx as f64;
    let utilization = total_usage / network_capacity as f64;
    
    // Sigmoid function to smooth the score
    network_score_weight * (1.0 / (1.0 + (-5.0 * (1.0 - utilization)).exp()))
}
```

atoma-state/src/handlers.rs
```diff
@@ -1767,3 +1908,1027 @@ mod utils {
...
+    /// let weight = 0.3;
+    ///
+    /// let score = compute_cpu_score(cpu_usage, weight);
+    /// // Returns 0.15 because:
+    /// ```
+    ///
+    /// # Notes
+    ///
+    /// - Returns 0.0 if the input is 100% CPU usage
+    fn compute_cpu_score(cpu_usage: f64, cpu_score_weight: f64) -> f64 {
```
maschad, line 2409: Hmmm I may be naive here but is the `cpu_usage` here considering all the cores a CPU might have? If not we should e.g.

```rust
fn compute_cpu_score(
    cpu_usage: f64,
    num_cores: i32,
    current_frequency: f64,
    max_frequency: f64,
    cpu_score_weight: f64
) -> f64 {
    let frequency_ratio = current_frequency / max_frequency;
    let core_utilization = cpu_usage / num_cores as f64;
    
    cpu_score_weight * (1.0 - core_utilization) * frequency_ratio
}
```

atoma-state/src/handlers.rs
```diff
@@ -1550,6 +1613,76 @@ pub(crate) async fn handle_node_metrics_registration_event(
...
+             - {} GPU power usages",
+            gpu_memory_used.len(),
+            gpu_memory_total.len(),
+            gpu_memory_free.len(),
+            gpu_percentage_time_read_write.len(),
+            gpu_percentage_time_execution.len(),
+            gpu_temperatures.len(),
+            gpu_power_usages.len()
+        );
+        return Err(AtomaStateManagerError::InvalidGpuMetrics(format!(
```
maschad, line 1645: We probably want to continue the trend of having granular errors as we do in other instances

e.g.

```rust
#[derive(Debug)]
enum PerformanceError {
    NoGpusAvailable,
    InvalidMetrics(String),
    DatabaseError(String),
    CalculationError(String),
}
```

maschad (changes requested):
Excellent work here @jorgeantonio21 this is some critical work 🚀  🥳  , and I am sure we may have to adjust this once it's integrated based on observations of the scores.

I've left some comments and suggestions.

I also think it would suit us to have some additional property-based tests for scoring functions e.g.

```rust
proptest! {
    #[test]
    fn test_gpu_vram_score_properties(
        free in vec(0i64..32_000_000_000i64, 1..4),
        total in vec(32_000_000_000i64..64_000_000_000i64, 1..4)
    ) {
        let score = compute_gpu_vram_utilization_score(&free, &total, 0.3);
        prop_assert!(score >= -1.0 && score <= 0.3);
    }

    #[test]
    fn test_gpu_temp_score_properties(temps in vec(20.0f64..100.0f64, 1..4)) {
        let score = compute_gpu_temp_score(&temps, 80.0, 100.0, 0.2);
        prop_assert!(score >= -1.0 && score <= 0.2);
    }
}
```

and also some integration tests for complete scoring pipeline e.g

```rust
#[tokio::test]
async fn test_performance_scoring_pipeline() -> Result<()> {
    let state = setup_test_db().await;
    let metrics = NodeMetrics {
        node_small_id: 1,
        cpu_usage: 0.5,
        ram_used: 8_000_000_000,
        ram_total: 16_000_000_000,
        gpu_memory_free: vec![8_000_000_000],
        gpu_memory_total: vec![16_000_000_000],
        gpu_percentage_time_execution: vec![0.3],
        gpu_temperatures: vec![70.0],
        gpu_power_usages: vec![180.0],
    };

    let performance = get_node_performance(&state, metrics).await?;
    
    // Test score components and correlations
    assert!(performance.total_performance_score >= 0.0 && performance.total_performance_score <= 1.0);
    
    // Test score changes with metrics changes
    let worse_metrics = NodeMetrics {
        cpu_usage: 0.9,
        ..metrics
    };
    let worse_performance = get_node_performance(&state, worse_metrics).await?;
    assert!(worse_performance.total_performance_score < performance.total_performance_score);

    Ok(())
}
```

Would love to hear your thoughts.

atoma-state/src/handlers.rs
jorgeantonio21: The calculation is definitely model dependent, as each model as different KV cache sizes, etc. For now, we are not collecting these metrics, so we will need to add it later. But it is definitely very important to move towards a KV cache VRAM analysis

atoma-state/src/handlers.rs
jorgeantonio21: Yeah, it must be GPU specific otherwise the formula does not make sense. This will be for a future work to have a better temperature analysis. For example, for a 8xH200 node the Max Thermal Design Power is 700W, however we are not collecting rn these more fine grained metrics around GPU model and node specs.

atoma-state/src/handlers.rs
jorgeantonio21: For sure, but we are not collecting `network_capacity` as of now. We will need to do it in the future

atoma-state/src/handlers.rs
jorgeantonio21: See https://github.com/atoma-network/atoma-node/issues/410

atoma-state/src/handlers.rs
jorgeantonio21: See https://github.com/atoma-network/atoma-node/issues/410

jorgeantonio21: And #232

atoma-state/src/handlers.rs
jorgeantonio21: And #232

atoma-state/src/handlers.rs
jorgeantonio21: And #232

atoma-state/src/handlers.rs
jorgeantonio21: I am not sure if I can collect the `current_frequency`. As far as I understand the frequency measures how many cycles a CPU can execute per second, and therefore this is a single value, per se. That said, it is a good number to capture, too.

atoma-state/src/handlers.rs
jorgeantonio21: Yeah, totally in favor of this

maschad: https://github.com/atoma-network/atoma-proxy/issues/233

maschad (approved):
Amazing work @jorgeantonio21

PR #224, opened 2025-02-12
feat: add client message to internal error

Cifko:
Added optional client message to internal error.
Please double check the filled message, so we don't accidentaly sends something that we shouldn't

### DIFF
--- a/atoma-proxy/src/server/handlers/chat_completions.rs
+++ b/atoma-proxy/src/server/handlers/chat_completions.rs
@@ -148,6 +148,7 @@ pub async fn chat_completions_create(
     .await
     .map_err(|e| AtomaProxyError::InternalError {
         message: format!("Failed to spawn image generation task: {e:?}"),
+        client_message: None,
         endpoint,
     })?
 }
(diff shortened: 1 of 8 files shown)


atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -692,6 +700,7 @@ async fn handle_streaming_response(
         .await
         .map_err(|e| AtomaProxyError::InternalError {
             message: format!("Error sending request to inference service: {e:?}"),
+            client_message: Some("Failed to connect to the node.".to_string()),
```
jorgeantonio21, line 703: Should we send the reason back to the client ? Or can we add more contex t?

jorgeantonio21 (approved):
LGTM

atoma-proxy/src/server/handlers/chat_completions.rs
Cifko: You want to send the actual error? I mean the connection failed (dropped, timed out, etc.) I don't think they need to know that. It's just there is something wrong with the connection.

PR #231, opened 2025-02-13
feat: add grafana dashboards

Cifko:
Add grafana dashboard methods.
This uses grafana api using private api key it grabs all the dashboard with a tag provided and forward all the dashboards/panels to the caller (In our case atoma-dashboard)


atoma-proxy-service/docs/openapi.yml
```diff
@@ -583,6 +583,30 @@ paths:
...
+      - Stats
+      summary: Get graphs.
+      description: |-
+        # Arguments
+
+        * `proxy_service_state` - The shared state containing the state manager
+
+        # Returns
+
+        * `Result<Json<Value>>` - A JSON response containing a list of graphs with data
```
maschad, line 598: Just to confirm, this is all external graphs correct?

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// # Arguments
+    ///
+    /// * `dashboard_uid` - The UID of the dashboard to get
+    ///
+    /// # Returns
+    ///
+    /// The dashboard with the specified UID
+    #[instrument(level = "info", skip(self))]
+    pub async fn get_dashboard(&self, dashboard_uid: String) -> Result<Dashboard, GrafanaError> {
+        let client = Client::new();
```
maschad, line 143: I don't think we need to create a client for each request, perhaps we could store a single client instance in the `Grafana` struct e.g.

```rust
pub struct Grafana {
    url: String,
    api_token: String,
    dashboard_tag: String,
    client: Client,  // Add this field
}

impl Grafana {
    pub fn new(url: String, api_token: String, dashboard_tag: String) -> Self {
        Self {
            url,
            api_token,
            dashboard_tag,
            client: Client::new(),
        }
    }
}
```

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+            .header("Accept", "application/json")
+            .header("Authorization", format!("Bearer {}", self.api_token))
+            .send()
+            .await?;
+
+        if response.status().is_success() {
+            let dashboard = response.json::<Dashboard>().await?;
+            Ok(dashboard)
+        } else {
+            Err(GrafanaError::FailedRequest(
```
maschad, line 156: We could perhaps make these errors a bit more granular as we do in other instances.

e.g.

```rust
pub enum GrafanaError {
    #[error("Request failed: {0}")]
    FailedRequest(#[from] reqwest::Error),
    #[error("Authentication failed")]
    AuthenticationError,
    #[error("Dashboard not found: {0}")]
    DashboardNotFound(String),
    #[error("Invalid response format: {0}")]
    InvalidResponse(#[from] serde_json::Error),
}
```

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+        dashboard: Dashboard,
+    ) -> Result<Vec<(String, Value)>, GrafanaError> {
+        let client = Client::new();
+        let request_url = format!("{}/api/ds/query", self.url);
+        let queries: Vec<(String, Query)> = dashboard.into();
+        let mut results = Vec::new();
+        for (title, query) in queries {
+            let response = client
+                .post(&request_url)
+                .header("Accept", "application/json")
```
maschad, line 183: There's some repetition in adding a header to each request, this could be extracted into a helper method

e.g.

```rust
impl Grafana {
    fn prepare_request(&self, builder: reqwest::RequestBuilder) -> reqwest::RequestBuilder {
        builder
            .header("Accept", "application/json")
            .header("Authorization", format!("Bearer {}", self.api_token))
            .header("Content-Type", "application/json")
    }
}
```

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// The tag to use to filter dashboards
+    dashboard_tag: String,
+}
+
+#[derive(Deserialize)]
+struct DashboardList {
+    uid: String,
+}
+
+#[derive(Deserialize)]
```
maschad, line 25: Any reason you chose to use `String` here over `chrono::DateTime` ?

```rust
#[derive(Deserialize)]
struct Time {
    from: DateTime<Utc>,
    to: DateTime<Utc>,
}
```

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `dashboard` - The dashboard to query data from
+    ///
+    /// # Returns
+    ///
+    /// A vector of tuples containing the title of the panel and the data
+    #[instrument(level = "info", skip_all)]
+    pub async fn query_data(
```
maschad, line 172: As we add more panels and they contain more data, processing them sequentially may not performant, perhaps we could stream them concurrently using futures.

e.g. 

```rust
use futures::stream::{self, StreamExt};

pub async fn query_data(&self, dashboard: Dashboard) -> Result<Vec<(String, Value)>, GrafanaError> {
    let queries: Vec<(String, Query)> = dashboard.into();
    
    // Create a stream from our queries
    stream::iter(queries)
        // Transform each query into a future
        .map(|(title, query)| async move {
            let response = self.client
                .post(&self.request_url)
                .json(&query)
                .send()
                .await?;
            
            let data = response.json().await?;
            Ok((title, data))
        })
        // Process multiple queries concurrently, we can configure this in the impl
        .buffer_unordered(self.concurrent_requests)  
        // Collect results into a vector
        .collect::<Vec<_>>()
        .await
}
```

maschad (changes requested):
Awesome job on this @Cifko 🚀 

I've left some suggestions and questions

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+impl Dashboard {
+    pub fn title(&self) -> String {
+        self.dashboard.title.clone()
+    }
+}
+
+#[derive(Debug, Serialize)]
+struct Query {
```
jorgeantonio21, line 56: Add docs

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+
+#[derive(Deserialize)]
+struct InnerDashboard {
+    panels: Vec<Panel>,
+    time: Time,
+    title: String,
+}
+
+#[derive(Deserialize)]
+pub struct Dashboard {
```
jorgeantonio21, line 45: Docs

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+#[derive(Deserialize)]
+struct Panel {
+    targets: Value,
+    title: String,
+}
+
+#[derive(Deserialize)]
+struct InnerDashboard {
```
jorgeantonio21, line 38: Docs

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+#[derive(Deserialize)]
+struct Time {
+    from: String,
+    to: String,
+}
+
+#[derive(Deserialize)]
+struct Panel {
```
jorgeantonio21, line 32: Docs

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// The URL of the Grafana instance
+    url: String,
+    /// The API token to use for authentication
+    api_token: String,
+    /// The tag to use to filter dashboards
+    dashboard_tag: String,
+}
+
+#[derive(Deserialize)]
+struct DashboardList {
```
jorgeantonio21, line 21: Docs

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+struct Query {
+    queries: Value,
+    from: String,
+    to: String,
+}
+
+/// Convert a `Dashboard` into a vector of queries
+impl From<Dashboard> for Vec<(String, Query)> {
+    fn from(dashboard: Dashboard) -> Self {
+        let from = dashboard.dashboard.time.from;
```
jorgeantonio21, line 65: You can directly destructure it like:

```rs
let Time { from, to } = dashboard.dashboard.time;
```

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+
+#[derive(Debug, Serialize)]
+struct Query {
+    queries: Value,
+    from: String,
+    to: String,
+}
+
+/// Convert a `Dashboard` into a vector of queries
+impl From<Dashboard> for Vec<(String, Query)> {
```
jorgeantonio21, line 63: Maybe best using a `HashMap`?

atoma-proxy-service/src/components/grafana.rs
```diff
@@ -0,0 +1,207 @@
...
+                        from: from.clone(),
+                        to: to.clone(),
+                    },
+                )
+            })
+            .collect()
+    }
+}
+
+impl Grafana {
```
jorgeantonio21, line 85: nitty: As a matter of taste, I usually prefer implementation bodies to be close to the struct declaration

atoma-proxy-service/src/components/grafana.rs
jorgeantonio21: We might want to do this at regular time intervals

atoma-proxy-service/src/components/grafana.rs
jorgeantonio21: Or process all headers at once

config.example.toml
```diff
@@ -27,6 +27,9 @@ hf_token = "<API_KEY>" # Hugging face api token, required if you want to access
 
 [atoma_proxy_service]
 service_bind_address = "0.0.0.0:8081" # This should match the port in the docker-compose.yml
+grafana_url = "" # Grafana URL
```
jorgeantonio21, line 30: This will panic for local instances of the proxy, I believe

atoma-proxy-service/src/handlers/stats.rs
```diff
@@ -275,3 +279,57 @@ async fn get_nodes_distribution(
...
+#[utoipa::path(
+    get,
+    path = "",
+    responses(
+        (status = OK, description = "Retrieves all graphs", body = Value),
+        (status = INTERNAL_SERVER_ERROR, description = "Failed to get graphs")
+    )
+)]
+#[instrument(level = "trace", skip_all)]
+async fn get_graphs(
```
jorgeantonio21, line 312: Don't we need some form of authorization here ? Or this being widely available is fine ?

jorgeantonio21 (commented):
Looking great !

atoma-proxy-service/docs/openapi.yml
Cifko: No, that's for graphs tagged with the tag that's specified in the config.
Weirdly, when you have grafana token you can't read the shared dashboards (you can list them, but not read them).
To be more precise, imagine you have dashboard A, it has some uid uia_a, then you share it. You can still read uid_a. But you can't read shared_uid_a. Which is different. You can probably find the mapping somewhere. 
But I've decided that if everyone is on the same note, that we use tags to share. We can even have different tags for different purpose in the future. Currently we have one tag and that's it.

atoma-proxy-service/src/components/grafana.rs
Cifko: Hashmap is slower when you don't do lookups. Also it has memory overhead compared to vec.
Since we do nothing of that sort Vec is better.
And also it would lead to problems when someone accidentally name the panels the same. Hashmap would just use the last of them.

atoma-proxy-service/src/components/grafana.rs
Cifko: Proxy don't want to do this at interval, the frontend should. Also we should enable caching at some point. Otherwise everyone who refreshes atoma-dashboard will do a bunch of grafana queries. Do you guys know if grafana has caching enabled by default? Probably not...
I'm of thinking of changing this to a different approach. To do this query_data request from the frontend. Because it's way to slow and we should probably just get the structure (names of the panels, # of them) first. So the frontend can show that with load on each graph.

atoma-proxy-service/src/components/grafana.rs
Cifko: Because the time is not actually time, it's `grafana time` which looks like `now` `now-1w` and similar. Also it's just passed around, it's not worth the compute time to do anything with it.

atoma-proxy-service/src/handlers/stats.rs
Cifko: If you don't want to hide the graphs on the dashboard behind login, then no.

atoma-proxy-service/src/components/grafana.rs
Cifko: I can do the status codes, but serde_json::Error is handled internally in the reqwest crate, I'm not going to convert it back

atoma-proxy-service/src/components/grafana.rs
Cifko: So I've change the structure. frontend get the name of the dashboard with panels, and then it call the query for each panel in parallel
This way the UI can show the graphs with loading information

config.example.toml
Cifko: why should it?
Just do http://localhost, and you are goot to go

atoma-proxy-service/docs/openapi.yml
maschad: That makes sense

jorgeantonio21 (approved):
LGTM

PR #244, opened 2025-02-19
build: add grafana server root url

maschad:


docker-compose.yaml
```diff
@@ -184,6 +184,8 @@ services:
     volumes:
       - grafana_data:/var/lib/grafana
     env_file: .env
+    environment:
+      - GF_SERVER_ROOT_URL=http://${GRAFANA_DOMAIN:-213.130.147.75}:${GRAFANA_PORT:-3000}/
```
jorgeantonio21, line 188: Why is `213.130.147.75` hardcoded ?

maschad: Similar to [here](https://github.com/atoma-network/atoma-node/pull/426#discussion_r1962225391) we can remove if we don't want it at all

jorgeantonio21: Yup, I think we don't want any explicit IP hardcoded in here

Issue #232, opened 2025-02-14
Compute better scoring for node performance


jorgeantonio21:
Following issue https://github.com/atoma-network/atoma-node/issues/410, we need to improve the node performance scoring computation, through by incorporating these more fine grained collected metrics. Ideally:

- [ ] Node specs are used to compute scores. These include KV cache allocations, thermal power values, cpu/gpu usage, bandwidth capacity.

These scores are used by the proxy to select the best nodes available to newly purchased `Stack`s (ahead of time). Then the proxy can use metrics around usage and performance, such as:

- [ ] Number of allocated KV blocks, node processed requests, throughput and latency (ITL and TTFT) across the board, to pick the most available node to process a new request (real time).

PR #252, opened 2025-02-23
feat: add gpu vllm metrics to the proxy

jorgeantonio21:
- Closes https://github.com/atoma-network/atoma-proxy/issues/232


atoma-state/src/timer.rs
```diff
@@ -0,0 +1,143 @@
...
+                                        NodeMetricsCollector::retrieve_best_available_nodes_for_image_generation(
+                                            &metrics_url,
+                                            modality_model,
+                                            top_k,
+                                        )
+                                        .await?
+                                    }
+                                };
+
+                                Ok::<_, MetricsServiceError>((*modality, best_available_nodes))
```
maschad, line 101: If we have an error in this main loop the service would be terminated, perhaps we should refactor this to a function 

```rust
async fn collect_and_send_metrics(
    config: &MetricsCollectionConfig,
    models: &[(Modalities, String)],
    tx: &mpsc::UnboundedSender<Vec<(Modalities, Vec<i64>)>>,
) -> Result<()> {
    let futures = models
        .iter()
        .map(|(modality, modality_model)| {
            let metrics_url = config.metrics_url.clone();
            async move {
                let best_available_nodes = match modality {
                    Modalities::ChatCompletions => {
                        NodeMetricsCollector::retrieve_best_available_nodes_for_chat_completions(
                            &metrics_url,
                            modality_model,
                            config.top_k,
                        )
                        .await?
                    }
                    Modalities::Embeddings => {
                        NodeMetricsCollector::retrieve_best_available_nodes_for_embeddings(
                            &metrics_url,
                            modality_model,
                            config.top_k,
                        )
                        .await?
                    }
                    Modalities::ImageGeneration => {
                        NodeMetricsCollector::retrieve_best_available_nodes_for_image_generation(
                            &metrics_url,
                            modality_model,
                            config.top_k,
                        )
                        .await?
                    }
                };

                Ok::<_, MetricsServiceError>((*modality, best_available_nodes))
            }
        });

    let best_available_nodes = futures::future::try_join_all(futures).await?;
    tx.send(best_available_nodes)
        .map_err(|_| MetricsServiceError::ChannelSendError)?;
    Ok(())
}
```

and then here in the main loop

```rust
            tokio::select! {
                _ = tokio::time::sleep(config.collection_interval) => {
                    if let Err(e) = collect_and_send_metrics(&config, &models, &tx_best_available_nodes).await {
                        tracing::error!(
                            target = "atoma-state-manager",
                            event = "metrics_collection_error",
                            error = %e,
                            "Failed to collect or send metrics"
                        );
                    }
                }
                shutdown_signal_changed = shutdown_signal.changed() => {
                    match shutdown_signal_changed {
                        Ok(()) => {
                            if *shutdown_signal.borrow() {
                                tracing::trace!(
                                    target = "atoma-state-manager",
                                    event = "shutdown_signal",
                                    "Shutdown signal received, shutting down"
                                );
                                break;
                            }
                        }
                        Err(e) => {
                            tracing::error!(
                                target = "atoma-state-manager",
                                event = "shutdown_signal_error",
                                error = %e,
                                "Shutdown signal channel closed"
                            );
                            break;
                        }
                    }
                }
            }
        }
        Ok(())
    })

```

atoma-state/src/metrics.rs
```diff
@@ -0,0 +1,707 @@
...
+};
+use tracing::instrument;
+
+use crate::{
+    errors::MetricsServiceError,
+    timer::{trigger_new_metrics_collection_task, Modalities},
+};
+
+/// Metrics timeout, in seconds
+const METRICS_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
```
maschad, line 21: Let's make this configurable

maschad (changes requested):
Great work on this @jorgeantonio21 🚀 

I recommend adding unit tests especially for error cases and Prometheus query parsing.

We should also seek to have https://github.com/atoma-network/atoma-node/pull/433 in first

atoma-state/src/config.rs
```diff
@@ -8,6 +8,9 @@ pub struct AtomaStateManagerConfig {
     /// The URL of the Postgres database.
     pub database_url: String,
 
+    /// The URL of the Prometheus metrics endpoint.
+    pub metrics_url: String,
```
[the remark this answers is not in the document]
jorgeantonio21: It is definitely

atoma-state/src/timer.rs
```diff
@@ -0,0 +1,143 @@
...
+    level = "debug",
+    skip_all,
+    fields(
+        metrics_url = %metrics_url,
+        models = ?models,
+        top_k = ?top_k
+    )
+)]
+pub fn trigger_new_metrics_collection_task(
+    metrics_url: String,
```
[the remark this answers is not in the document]
jorgeantonio21: Makes sense, refactored

atoma-state/src/timer.rs
jorgeantonio21: Great call here, addressed

atoma-state/src/metrics.rs
jorgeantonio21: What's the benefit ?

maschad: The main benefit is that we may notice that metrics are slower in being scraped that first anticipated, and rather than having to do a programmatic change we can change a config and re-deploy.

jorgeantonio21: That is true, but ideally any deployed proxy will follow the default value, unless having a very good reason to not do so. But I can add it to the config

atoma-state/src/metrics.rs
```diff
@@ -0,0 +1,707 @@
...
+        )
+    )]
+    pub async fn retrieve_best_available_nodes_for_chat_completions(
+        metrics_url: &str,
+        model: &str,
+        top_k: Option<usize>,
+    ) -> Result<Vec<i64>> {
+        let top_k = top_k.unwrap_or(DEFAULT_TOP_K_BEST_AVAILABLE_NODES);
+        let query = format!(
+            r#"topk({top_k},
```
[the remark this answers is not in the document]
jorgeantonio21: This has been addressed

atoma-state/src/metrics.rs
```diff
@@ -0,0 +1,707 @@
+use std::collections::HashMap;
+
+use atoma_p2p::metrics::{
+    ChatCompletionsMetrics, EmbeddingsMetrics, ImageGenerationMetrics, ModelMetrics, NodeMetrics,
+};
+use once_cell::sync::Lazy;
+use prometheus::{GaugeVec, Opts, Registry};
```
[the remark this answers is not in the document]
maschad: I've decided against this for a few reasons, primarily there isn't an easy way to reset metrics in opentelemetry and also `GaugeVec` suits our use case well, which doesn't exist in opentelemetery

maschad (approved):

PR #261, opened 2025-03-03
feat: add metrics collection configuration to example config

Cifko:
Add metrics collection configuration to example config


jorgeantonio21 (approved):
LGTM

PR #268, opened 2025-03-04
feat: update config.example.toml and remove debug exporter from otel-collector

jorgeantonio21:


README.md
```diff
@@ -76,7 +76,13 @@ git clone https://github.com/atoma-network/atoma-proxy.git
 cd atoma-proxy
 ```
 
-1. Configure environment variables by creating `.env` file. Please ensure you have created the requisite user and database in your postgres instance. Once you've done that, you can use `.env.example` as template for your `.env` file.
+2. Configure environment variables by creating `.env` file, with:
```
maschad, line 79: This suggests that merely copying the example to `.env` is sufficient, when really a user ought to populate it with their specific information. I think the original captured that better.

maschad (changes requested):

README.md
jorgeantonio21: Addressed

PR #274, opened 2025-03-04
feat: map cursor toml file to host dir on docker compose file

jorgeantonio21:


maschad (changes requested):
If the intention is to persist the `cursor.toml` file you will need to map it to a [volume](https://github.com/atoma-network/atoma-proxy/pull/274/files#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9R224)

maschad (approved):
Upon reconsideration this volume mapping is not explicitly required

PR #278, opened 2025-03-05
fix: add logs for loki

maschad:


atoma-proxy/src/telemetry.rs
```diff
@@ -78,13 +80,24 @@ fn init_traces(otlp_endpoint: &str) -> Result<sdktrace::Tracer> {
...
     global::set_meter_provider(metrics_provider);
 
+    let (layer, task) = tracing_loki::builder()
+        .label("service_name", "atoma-proxy")?
+        .extra_field("pid", format!("{}", process::id()))?
+        .build_url(Url::parse(&loki_endpoint).unwrap())?;
+
+    // The background task needs to be spawned so the logs actually get
+    // delivered.
+    tokio::spawn(task);
```
jorgeantonio21, line 99: In that case, we should keep the join handle for later appropriate resource cleaning

jorgeantonio21 (approved):
LGTM

PR #288, opened 2025-03-11
feat: return 0 balance when user is not found

Cifko:
Before the non existing user was handled in the frontend, let just return 0 for non-existing user here to simplify the frontend.
Non-existing user means the user had not yet topped up his account. But the users exists because this function is behind auth.


jorgeantonio21 (approved):
LGTM

PR #296, opened 2025-03-17
chore: add metrics for ITL and TTFT

maschad:


atoma-proxy/src/server/streamer.rs
```diff
@@ -340,6 +360,13 @@ impl Stream for Streamer {
 
                 let chunk = match serde_json::from_str::<Value>(chunk_str) {
                     Ok(chunk) => {
+                        if let Some(timer) = self.inter_stream_token_latency_timer.take() {
```
jorgeantonio21, line 363: I believe you need to set a new timer instant for the next token

jorgeantonio21 (commented):
Left a comment, otherwise looks good to me

atoma-proxy/src/server/streamer.rs
maschad: Good catch! added 3c1659a

PR #297, opened 2025-03-17
feat: add last used timestamp for tokens

Cifko:
Whenever the api tokens is checked for validity, we count it as used and modify the timestamp.


jorgeantonio21 (approved):
LGTM

maschad (approved):
LGTM

PR #335, opened 2025-03-26
feat: stack check for confidential compute

jorgeantonio21:


Cifko (approved):
This has no functional change.

PR #345, opened 2025-03-28
feat: acquire new stack in thread

Cifko:
Acquire new stack in a tokio thread to prevent killing the code between acquiring and storing the stack (that can happen when the connection is dropped).

### DIFF
--- a/atoma-proxy/src/server/middleware.rs
+++ b/atoma-proxy/src/server/middleware.rs
@@ -222,7 +222,8 @@ impl RequestMetadataExtension {
 #[instrument(
     level = "info",
     skip_all,
-    fields(endpoint = %req.uri().path())
+    fields(endpoint = %req.uri().path()),
+    err
 )]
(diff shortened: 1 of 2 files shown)


atoma-proxy/src/server/middleware.rs
```diff
@@ -869,6 +869,112 @@ pub mod auth {
...
+    /// #Arguments
+    ///
+    /// * `node` - The cheapest node to acquire a stack for
+    ///
+    /// #Returns
+    ///
+    /// Returns a `NewStackResult` containing:
+    /// * `stack_small_id` - The identifier for the selected/created stack
+    /// * `selected_node_id` - The identifier for the node that will process the request
+    #[instrument(level = "info", skip_all)]
```
jorgeantonio21, line 892: add err

jorgeantonio21 (approved):
LGTM

Left only a nitty comment

PR #350, opened 2025-03-28
feat: handle streamer drop openrouter

jorgeantonio21:


atoma-proxy/src/server/middleware.rs
```diff
@@ -596,7 +603,7 @@ pub mod auth {
         headers: &HeaderMap,
         body_json: &Value,
         endpoint: &str,
-    ) -> Result<(Option<Stack>, u64, String, i64)> {
+    ) -> Result<(Option<Stack>, u64, u64, String, i64)> {
```
Cifko, line 606: this tuple is getting out of hand, change it to struct

atoma-proxy/src/server/middleware.rs
```diff
@@ -708,14 +715,17 @@ pub mod auth {
         headers: &HeaderMap,
         request_model: impl RequestModel + Send,
         endpoint: &str,
-    ) -> Result<(Option<Stack>, u64, String, i64)> {
+    ) -> Result<(Option<Stack>, u64, u64, String, i64)> {
```
Cifko, line 718: same tuple as above, right?

Cifko (approved):

atoma-proxy/src/server/middleware.rs
jorgeantonio21: Addressed, thanks for pointing it out

PR #352, opened 2025-03-29
feat: code cleanup, remove unnecessary code

jorgeantonio21:
@Cifko please make sure none of these removal are necessary for graphics, etc.


Cifko (commented):
The latency is used, getstatsstacks is not.
But maybe we should change it to grafana
It's these we still compute ourselfs 
image

jorgeantonio21:
Alright, so shall we keep this PR open until we fully migrate to Grafana ? @Cifko

PR #353, opened 2025-03-29
feat: add logic for acquire stack lock

jorgeantonio21:
@Cifko we need to review this one very carefully, and test it throughly.

I have introduced a new struct `LockGuard` to ensure that the any acquire stack lock encapsules the entire logic of acquiring a new stack (and there are no possible race conditions between logical branches, otherwise).

This `LockGuard` also implements the `Drop` trait making sure that resources are cleanup appropriately. We should do the same for the `DashMap` use in the `atoma-node` for counting concurrent requests.

```rs
    /// A guard that locks a stack for a user id.
    ///
    /// This struct is used to lock a stack for a user id, so that no other concurrent requests can try to acquire a new stack,
    /// to avoid buying multiple redundant stacks, at the same time (that is in a window of 300ms, following Sui's Mysticeti fast finality estimation times).
    pub struct LockGuard<'a> {
        /// The map of user id to lock status.
        map: &'a DashMap<i64, bool>,
        /// The user id to lock.
        key: i64,
        /// Whether the lock is held.
        locked: bool,
    }
    impl<'a> LockGuard<'a> {
        /// Tries to lock a stack for a user id.
        ///
        /// This method checks if the user id is already locked in the map. If it is, it returns `None`,
        /// indicating that the stack is already locked. Otherwise, it locks the stack and returns a new `LockGuard`.
        ///
        /// # Arguments
        ///
        /// * `map` - The map of user id to lock status.
        /// * `key` - The user id to lock.
        ///
        /// # Returns
        ///
        /// Returns a new `LockGuard` if the stack is not locked, otherwise `None`.
        pub fn try_lock(map: &'a DashMap<i64, bool>, key: i64) -> Option<Self> {
            match map.entry(key) {
                dashmap::mapref::entry::Entry::Occupied(_) => None, // Already locked
                dashmap::mapref::entry::Entry::Vacant(entry) => {
                    entry.insert(true);
                    Some(Self {
                        map,
                        key,
                        locked: true,
                    })
                }
            }
        }
    }

    impl Drop for LockGuard<'_> {
        /// Drops the lock guard and releases the lock.
        ///
        /// This method removes the lock from the map when the lock guard goes out of scope.
        /// It logs a message indicating that the lock has been released for the user id.
        #[instrument(level = "info", skip_all, fields(user_id = %self.key))]
        fn drop(&mut self) {
            if self.locked {
                let previous_lock = self.map.remove(&self.key);
                if previous_lock.is_some() {
                    info!("Held lock has been released for user id: {}", self.key);
                }
            }
        }
    }
```

### DIFF
--- a/atoma-proxy/src/server/middleware.rs
+++ b/atoma-proxy/src/server/middleware.rs
@@ -599,6 +599,7 @@ pub mod auth {
         http_server::ProxyState, Result, ONE_MILLION,
     };
 
+    use super::acquire_stack_lock;
     use super::STACK_SIZE_TO_BUY;
 
     /// The maximum time to wait for a stack to be created, on the Sui blockchain,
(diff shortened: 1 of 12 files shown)


atoma-proxy/src/server/middleware.rs
```diff
@@ -1081,6 +1187,166 @@ pub mod auth {
...
+            }
+        }
+        Err(AtomaProxyError::RequestError {
+            message: format!("User {user_id} has already bought a stack"),
+            endpoint: endpoint.to_string(),
+        })
+    }
+
+    /// Arguments for the get_selected_node function.
+    pub struct GetSelectedNodeArgs<'a> {
```
Cifko, line 1331: Why creating a struct to pass to a function, can't we just pass the arguments?

Cifko (changes requested):
I stopped in the middle of review, because this is not good. If users wants to buy stacks for two different model, it should be allowed, and it's not, the dashmap is for user true/false, it should be user,task_id,true/false map...

atoma-proxy/src/server/middleware.rs
jorgeantonio21: Let's pass the struct, otherwise too many arguments, and I get lost passing them around in order

atoma-proxy/src/server/middleware.rs
```diff
@@ -1081,6 +1187,188 @@ pub mod auth {
...
+        if stack_is_locked {
+            // NOTE: This means a concurrent request is already buying a stack, so we wait for it to finish,
+            // and for the stack creation event to be stored on the AtomaStateManager's internal state, as
+            // we store it right after the call to the Sui blockchain and don't wait to catch the event
+            // from the Sui blockchain, as this can take extra time (roughly 300ms following Sui's Mysticeti
+            // finality times). We will try again, for a fixed number of times. If the stack is still not created,
+            // we return an error.
+            for _ in 0..MAX_STACK_WAIT_ATTEMPTS {
+                let stack_metadata =
+                    try_get_stack_for_user_id(state, user_id, request_model.clone(), endpoint)
```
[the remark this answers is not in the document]
jorgeantonio21: Actually, in theory, it is possible that these are different..

Cifko (approved):

PR #373, opened 2025-04-01
fix: negative tokens

Cifko:
Fix negative token when waiting for acquire new stack


jorgeantonio21 (approved):
Lgtm

Update the docs of the state manager method

PR #380, opened 2025-04-02
feat: handle too early requests

jorgeantonio21:
Due to overestimation of compute units, a node might not have enough available compute units for a stack and send a lock message to the proxy, in scenarios of multiple concurrent requests. This is not ideal, as the stack might not be locked (and actually far from it), so there is not reason for the proxy to locked it in its internal state.

Instead, we introduce a new status code `TOO_EARLY`, meaning that this request needs either to be retried at a moment's later (confidential endpoint) or the proxy just acquires a new stack for that user's and model/task, so that it then can handle such large volumes of concurrent requests at the same time.

Confidential endpoints requires a full retrial (no stack to be bought by the proxy), as the request was encrypted with a specific stack assigned node's public encryption key. So instead, the logic should be on the Atoma's SDKs to retry for a few times.

### DIFF
--- a/atoma-proxy/src/server/middleware.rs
+++ b/atoma-proxy/src/server/middleware.rs
@@ -1656,7 +1656,7 @@ pub mod auth {
             }
         }
         Err(AtomaProxyError::RequestError {
-            message: format!("User {user_id} has already bought a stack"),
+            message: "Many concurrent requests, a stack is being bought, but its internal state is not yet updated. Please retry.".to_string(),
             endpoint: endpoint.to_string(),
         })
     }
--- a/atoma-proxy/src/server/middleware.rs
+++ b/atoma-proxy/src/server/middleware.rs
@@ -1860,7 +1860,7 @@ pub mod auth {
             lock_guard,
             endpoint.to_string(),
             max_total_num_compute_units,
-            Arc::clone(state.sui),
+            Arc::clone(&state.sui),
             node,
         )
         .await
@@ -2027,6 +2027,7 @@ pub mod utils {
             num_input_tokens: Some(num_input_tokens),
             num_compute_units: total_compute_units,
             selected_stack_small_id,
+            user_id,
             endpoint: endpoint.to_string(),
             model_name: request_model.to_string(),
         });
--- a/atoma-proxy/src/server/middleware.rs
+++ b/atoma-proxy/src/server/middleware.rs
@@ -63,6 +63,9 @@ pub struct RequestMetadataExtension {
     /// the input tokens are not known, so we don't count it.
     pub num_input_tokens: Option<u64>,
 
+    /// The user id for this request.
+    pub user_id: i64,
+
     /// Estimated compute units required for this request.
     /// This represents the total computational resources needed for both input and output processing.
     pub num_compute_units: u64,
@@ -123,6 +126,11 @@ impl RequestMetadataExtension {
     /// Returns self with the num compute units field populated, enabling method chaining
     pub const fn with_num_compute_units(mut self, num_compute_units: u64) -> Self {
         self.num_compute_units = num_compute_units;
+        self
+    }
+
+    pub const fn with_user_id(mut self, user_id: i64) -> Self {
+        self.user_id = user_id;
         self
     }
 
@@ -413,6 +421,7 @@ pub async fn confidential_compute_middleware(
 ) -> Result<Response> {
     let (mut req_parts, body) = req.into_parts();
     let endpoint = req_parts.uri.path().to_string();
+    let user_id = check_auth(&state.state_manager_sender, &req_parts.headers, &endpoint).await?;
     let body_bytes = axum::body::to_bytes(body, MAX_BODY_SIZE)
         .await
         .map_err(|e| {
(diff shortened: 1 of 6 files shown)


atoma-proxy/src/server/middleware.rs
```diff
@@ -587,20 +613,103 @@ pub async fn handle_locked_stack_middleware(
...
+                .headers
+                .insert(AUTHORIZATION, authorization_header);
+            let req = Request::from_parts(req_parts, Body::from(body_bytes));
+            if is_confidential_compute_endpoint(&endpoint) {
+                // NOTE: If this is a confidential compute request, the client already sent a request encrypted for a specific node.
+                // Acquiring a new stack in this case might simply select a new node, which means that the client will have to re-encrypt the request,
+                // with the new node's public key. For this reason, instead of being the proxy retrying to acquire a new stack, we let the client retry
+                // a couple of times, through Atoma's native SDKs.
+                return Err(AtomaProxyError::Locked {
+                    message: "Confidential compute requests should be retried".to_string(),
```
maschad, line 630: Let's change the error code here to be more specific to the `LOCKED` scenario.

atoma-proxy/src/server/middleware.rs
```diff
@@ -587,20 +613,103 @@ pub async fn handle_locked_stack_middleware(
...
-    } else {
-        Ok(response)
+        StatusCode::TOO_EARLY => {
+            // NOTE: In this case, the node hasn't locked the stack immediately (has overestimated the compute units required for the request).
+            if is_confidential_compute_endpoint(&endpoint) {
+                // NOTE: If this is a confidential compute request, the client already sent a request encrypted for a specific node.
+                // Acquiring a new stack in this case might simply select a new node, which means that the client will have to re-encrypt the request,
+                // with the new node's public key. For this reason, instead of being the proxy retrying to acquire a new stack, we let the client retry
+                // a couple of times, through Atoma's native SDKs.
+                return Err(AtomaProxyError::UnavailableStack {
```
maschad, line 645: For the `TOO_EARLY` case, you're using `UnavailableStack` for confidential compute but don't have a specific error type for the same condition in the `LOCKED` case (using `AtomaProxyError::Locked` instead). Consistent error types would make client-side error handling more straightforward.

atoma-proxy/src/server/middleware.rs
```diff
@@ -587,20 +613,103 @@ pub async fn handle_locked_stack_middleware(
...
+                    endpoint: endpoint.to_string(),
+                });
+            }
+            // NOTE: For non-confidential requests, we can acquire a new stack directly from the proxy, and therefore we require a new authentication
+            // for the request.
             authenticate_middleware(state, req, next).await
         }
-    } else {
-        Ok(response)
+        StatusCode::TOO_EARLY => {
```
maschad, line 638: We should add a telemetric counter here to track how often this is happening, likewise for the locked, to better inform resource estimation.

atoma-proxy/src/server/middleware.rs
```diff
@@ -587,20 +613,103 @@ pub async fn handle_locked_stack_middleware(
...
+                // NOTE: If this is a confidential compute request, the client already sent a request encrypted for a specific node.
+                // Acquiring a new stack in this case might simply select a new node, which means that the client will have to re-encrypt the request,
+                // with the new node's public key. For this reason, instead of being the proxy retrying to acquire a new stack, we let the client retry
+                // a couple of times, through Atoma's native SDKs.
+                return Err(AtomaProxyError::UnavailableStack {
+                    message: "Confidential compute requests should be retried".to_string(),
+                    endpoint: endpoint.to_string(),
+                });
+            }
+            // We need to acquire a new stack for the request, to be able to retry
```
maschad, line 650: Would we do a similar re-try when the resources are locked as well?

maschad (changes requested):
Nice work @jorgeantonio21 🚀  left some comments

atoma-proxy/src/server/middleware.rs
jorgeantonio21: Yes we should have different status codes for different scenarios. Treating both scenarios the same would be incorrect

atoma-proxy/src/server/middleware.rs
jorgeantonio21: That's a great point ! will add that

atoma-proxy/src/server/middleware.rs
jorgeantonio21: You mean the message ?

maschad: Sorry yes I meant the message, both the `LOCKED` and `TOO_EARLY` say `Confidential compute requests should be retried`

atoma-proxy/src/server/middleware.rs
jorgeantonio21: In the case of a stack is locked, the proxy will try to acquire a new stack for you. In the confidential compute case it can't (because there is no guarantee the newly chosen node will be the same to which requests were encrypted in the first place).

Cifko (approved):

PR #382, opened 2025-04-03
feat: add locked compute units

Cifko:
Add locked compute units to stacks table.

### DIFF
--- a/atoma-state/src/types.rs
+++ b/atoma-state/src/types.rs
@@ -430,7 +430,7 @@ pub struct Stack {
     /// Number of compute units already processed
     pub already_computed_units: i64,
 
-    /// Number of compute units already processed
+    /// Number of locked compute units for processing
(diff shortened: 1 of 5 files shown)


atoma-state/src/types.rs
```diff
@@ -430,6 +430,9 @@ pub struct Stack {
     /// Number of compute units already processed
     pub already_computed_units: i64,
 
+    /// Number of compute units already processed
```
jorgeantonio21, line 433: Exactly the same comment as for the previous field.

jorgeantonio21 (commented):
LGTM

Was this tested ? Btw, please address the comments above

Cifko:
> LGTM
> 
> Was this tested ? Btw, please address the comments above

Tested, the `locked_compute_units` always goes to zero and the `already_computed_units` matches the stack on the node.

PR #386, opened 2025-04-05
feat: check existing stack before acquiring

jorgeantonio21:


atoma-proxy/src/server/middleware.rs
```diff
@@ -1882,6 +1901,72 @@ pub mod auth {
...
+                result_sender,
+            })
+            .map_err(|err| AtomaProxyError::InternalError {
+                message: format!("Failed to send GetStacksForModel event: {err:?}"),
+                client_message: None,
+                endpoint: endpoint.to_string(),
+            })?;
+        let maybe_stack = result_receiver
+            .await
+            .map_err(|err| AtomaProxyError::InternalError {
```
maschad, line 1954: Why do we need this duplicate?

atoma-proxy/src/server/middleware.rs
```diff
@@ -687,16 +687,35 @@ pub async fn handle_locked_stack_middleware(
...
+                None => {
+                    // 2. Acquire a new stack for the request, this will also lock compute units for the new acquired stack
+                    get_cheapest_node_and_acquire_new_stack(
+                        &state,
+                        user_id,
+                        &request_metadata.model_name,
+                        &request_metadata.endpoint,
+                        max_total_num_compute_units,
+                    )
+                    .await?
```
maschad, line 715: We should probably handle this explicitly

atoma-proxy/src/server/middleware.rs
```diff
@@ -687,16 +687,35 @@ pub async fn handle_locked_stack_middleware(
             // We need to acquire a new stack for the request, to be able to retry
             let user_id = request_metadata.user_id;
             let max_total_num_compute_units = request_metadata.max_total_num_compute_units;
-            // 1. Acquire a new stack for the request, this will also lock compute units for the new acquired stack
-            let selected_node_metadata = get_cheapest_node_and_acquire_new_stack(
+            // 1. Try to get a Stack from the state manager
```
maschad, line 690: Do you think there could be a potential race condition here?

```rust
// Use a mutex or other synchronization mechanism
let stack_lock = state.stack_acquisition_mutex.lock().await;
let result = acquire_stack_with_retry(&state, user_id, &request_metadata).await;
drop(stack_lock); // Release the lock
```

atoma-proxy/src/server/middleware.rs
```diff
@@ -687,16 +687,35 @@ pub async fn handle_locked_stack_middleware(
...
+            let maybe_stack = get_node_metadata_from_state_manager(
```
maschad, line 691: We should add some `debug` logs here 

```rust
tracing::debug!(
    "Attempting to acquire compute stack for user {} with model {} and max units {}",
    request_metadata.user_id,
    request_metadata.model_name,
    max_total_num_compute_units
);

// Later
tracing::debug!(
    "Successfully acquired stack {} on node {} for user {}",
    selected_node_metadata.stack_small_id,
    selected_node_metadata.selected_node_id,
    user_id
);
```

maschad (changes requested):
Nice work @jorgeantonio21 left some comments

atoma-proxy/src/server/middleware.rs
jorgeantonio21: We have already handled this before. This code should be safe

jorgeantonio21: No, we have handled race conditions situations before.

atoma-proxy/src/server/middleware.rs
jorgeantonio21: How so ?

atoma-proxy/src/server/middleware.rs
jorgeantonio21: Because one result is about channel closed, the other result consists of possible db failures.

atoma-proxy/src/server/middleware.rs
jorgeantonio21: This operation is not related to stack acquistion, it instead tries to read from the db.

jorgeantonio21: I added a few of these logs at the right place.

atoma-proxy/src/server/middleware.rs
jorgeantonio21: Btw, the suggestion would not help in this case, as it would try to always buy stacks, instead of querying the db if a lock already exists on some other task.

atoma-proxy/src/server/middleware.rs
maschad: such as 

```rust
  match get_cheapest_node_and_acquire_new_stack(
                        &state,
                        user_id,
                        &request_metadata.model_name,
                        &request_metadata.endpoint,
                        max_total_num_compute_units,
                    )
                    .await
                    {
                        Ok(metadata) => metadata,
                        Err(e) => {
                            return Err(AtomaProxyError::InternalError {
                                message: format!("Unable to acquire new stack: {e}"),
                                client_message: Some(
                                    "Failed to acquire compute resources".to_string(),
                                ),
                                endpoint: request_metadata.endpoint.clone(),
                            })
                        }
                    }
  ```

Comment (jorgeantonio21) [reply atoma-proxy/src/server/middleware.rs:715]: That's already done implicitly through error propagation

File: atoma-proxy/src/server/middleware.rs [hunk 5]
```diff
@@ -687,16 +687,35 @@ pub async fn handle_locked_stack_middleware(
...
                 &state,
-                user_id,
                 &request_metadata.model_name,
-                &request_metadata.endpoint,
-                max_total_num_compute_units,
+                user_id,
+                max_total_num_compute_units as i64,
+                is_confidential_compute_endpoint(&endpoint),
```
Cifko, line 696: this is never confidential, because on line 677 you throw an error for confidential
so I just want to make sure, that you are aware of that

jorgeantonio21: It doesn't hurt to have the check in here too, and if we want to change the logic later on, it is more robust to have the check in place than if not

Cifko (approved):

PR #399, opened 2025-04-09
feat: add OpenRouter models support and configuration

Cifko:
Add new open router models endpoint.
It just copy the content of the json file. So we need to add proper file. The `open_router.json` is mapped to docker to provide the content.

### DIFF
--- a/atoma-proxy/src/server/handlers/models.rs
+++ b/atoma-proxy/src/server/handlers/models.rs
@@ -84,6 +84,15 @@ pub struct Model {
     pub owned_by: String,
 }
 
+/// OpenAPI documentation for the open router models listing endpoint.
+///
+/// This struct is used to generate OpenAPI documentation for the models listing
+/// endpoint. It uses the `utoipa` crate's derive macro to automatically generate
+/// the OpenAPI specification from the code.
+#[derive(OpenApi)]
+#[openapi(paths(open_router_models_list))]
+pub struct OpenRouterModelsListApi;
+
 /// OpenRouter models listing endpoint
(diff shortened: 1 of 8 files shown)


atoma-proxy/src/server/handlers/models.rs
```diff
@@ -77,3 +82,24 @@ pub struct Model {
     /// Organization that owns the model
     pub owned_by: String,
 }
+
+pub async fn open_router_models_list(
```
jorgeantonio21, line 86: Add docs and instrumentation

atoma-proxy/src/server/http_server.rs
```diff
@@ -223,7 +227,9 @@ pub fn create_router(state: &ProxyState) -> Router {
         .route(NODES_CREATE_PATH, post(nodes_create))
         .route(NODES_CREATE_LOCK_PATH, post(nodes_create_lock));
 
-    let public_routes = Router::new().route(HEALTH_PATH, get(health));
+    let public_routes = Router::new()
+        .route(HEALTH_PATH, get(health))
+        .route(OPEN_ROUTER_MODELS_PATH, get(open_router_models_list));
```
jorgeantonio21, line 232: No need to authenticate ?

jorgeantonio21 (commented):
LGTM

But it is missing the updates on openapi.yaml file

atoma-proxy/src/server/handlers/models.rs
Cifko: added

Issue #250, opened 2025-02-20
Track model usage per wallet address


maschad:
We want to see how a wallet might use its stack per model per service request

PR #400, opened 2025-04-09
feat: track model usage per wallet address

maschad:
- Closes https://github.com/atoma-network/atoma-proxy/issues/250


atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -169,11 +169,25 @@ pub async fn chat_completions_create(
...
         match handle_chat_completions_request(&state, &metadata, headers, payload, is_streaming)
             .await
         {
             Ok(response) => {
-                TOTAL_COMPLETED_REQUESTS.add(1, &[KeyValue::new("model", metadata.model_name)]);
+                TOTAL_COMPLETED_REQUESTS.add(
+                    1,
+                    &[
+                        KeyValue::new("model", metadata.model_name),
+                        KeyValue::new("wallet_address", wallet_address),
```
jorgeantonio21, line 188: This doesn't make that much sense, since the proxy wallet is static (never changes), so we shouldn't filter by its value. On the other hand, we should filter by model, as before

jorgeantonio21 (changes requested):
This PR is not really doing what was intended.

atoma-proxy/src/server/handlers/chat_completions.rs
maschad: Correct, my bad I conflated the proxy wallet with the user's wallet, which in fact is our `user_id`

maschad:
Closing in favour of https://github.com/atoma-network/atoma-proxy/pull/403

PR #401, opened 2025-04-10
feat: completions endpoint

Cifko:
Add completions endpoint, that redirects to chat/completions


atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -197,6 +205,75 @@ pub async fn chat_completions_create(
...
+            _ => {
+                return Err(AtomaProxyError::RequestError {
+                    message: "Invalid 'prompt' field".to_string(),
+                    endpoint: COMPLETIONS_PATH.to_string(),
+                });
+            }
+        };
+
+        // Replace "prompt" with "messages"
+        payload.as_object_mut().unwrap().remove("prompt");
```
jorgeantonio21, line 261: Don't use unwrap in here

Cifko: removed

PR #403, opened 2025-04-10
feat: review prometheus metrics

jorgeantonio21:
- Closes https://github.com/atoma-network/atoma-proxy/issues/250

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,7 +148,7 @@ dependencies = [
  "rcgen",
  "ring",
  "rustls",
- "rustls-webpki 0.103.0",
+ "rustls-webpki 0.103.1",
  "serde",
  "serde_json",
  "socket2",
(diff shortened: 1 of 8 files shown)


atoma-proxy/src/server/handlers/chat_completions.rs
```diff
@@ -178,6 +181,8 @@ pub async fn chat_completions_create(
             Err(e) => {
                 let model_label: String = metadata.model_name.clone();
                 TOTAL_FAILED_CHAT_REQUESTS.add(1, &[KeyValue::new("model", model_label)]);
```
maschad, line 183: Why not add a `user_id` label to the `TOTAL_FAILED_CHAT_REQUESTS` as opposed to creating a new metric for it?

maschad (changes requested):
Nice work here @jorgeantonio21 but is there any reason you opted to go the route of creating new metrics specifically for user related tracking as opposed to just adding a label? I am inclined to think that using this label allows us to leverage existing metrics.

jorgeantonio21:
@maschad I prefer having separated metrics per model and user id

maschad (approved):

PR #424, opened 2025-04-22
fix: store the txn digest during node selection

maschad:


atoma-proxy/src/server/handlers/nodes.rs
```diff
@@ -336,7 +337,7 @@ pub async fn nodes_create_lock(
             Ok(Json(NodesCreateLockResponse {
                 public_key,
                 node_small_id: node_public_key.node_small_id as u64,
-                stack_entry_digest: None,
+                stack_entry_digest: TransactionDigest::ZERO.base58_encode(),
```
jorgeantonio21, line 340: No, this is not good. It must be the actual exact transaction digest of the stack.

atoma-proxy/src/server/middleware.rs
```diff
@@ -357,7 +360,16 @@ pub async fn authenticate_middleware(
...
+            TransactionDigest::from_str(&tx_digest)
+                .map_err(|e| {
+                    tracing::error!("Failed to parse tx_digest: {e}");
+                    AtomaProxyError::InternalError {
+                        message: format!("Failed to parse tx_digest: {e}"),
+                        client_message: None,
+                        endpoint: endpoint.to_string(),
+                    }
+                })
+                .expect("Failed to parse tx_digest"),
```
jorgeantonio21, line 372: Why panicking ?

atoma-proxy/src/server/middleware.rs
```diff
@@ -1820,7 +1834,7 @@ pub mod auth {
             return Ok(SelectedNodeMetadata {
                 stack_small_id: stack.stack_small_id,
                 selected_node_id: stack.selected_node_id,
-                tx_digest: None,
+                tx_digest: stack.tx_digest,
```
jorgeantonio21, line 1837: It seems to me that sometime we return base58 encoded tx digest, sometimes not..

atoma-state/src/migrations/20250411093352_add-tx-digest-to-stacks.sql
```diff
@@ -0,0 +1,8 @@
+-- Add tx_digest column to stacks table
+ALTER TABLE
```
jorgeantonio21, line 2: This formatting looks a bit weird to me, can we follow the usual formatting ?

atoma-state/src/migrations/20250411093352_add-tx-digest-to-stacks.sql
```diff
@@ -0,0 +1,8 @@
...
+	stacks
+ADD
+	COLUMN tx_digest TEXT;
```
jorgeantonio21, line 5: Do we want to add a default value to be empty string ? and make it a required value

jorgeantonio21 (changes requested):
You are not storing the `tx_digest` in the stacks table.

atoma-proxy/src/server/middleware.rs
maschad: This is because upon first acquisition of a stack, the `tx_digest` is encoded as `base58` string and then stored in the db, subsequent retrievals such as this case, of that `tx_digest` would be from the db and therefore it would already be a `String` and would not be required to be re-encoded

atoma-state/src/migrations/20250411093352_add-tx-digest-to-stacks.sql
maschad: Good point done in 24c7dcf

atoma-state/src/migrations/20250411093352_add-tx-digest-to-stacks.sql
maschad: 24c7dcf

atoma-proxy/src/server/middleware.rs
maschad: Good catch 24c7dcf

atoma-proxy/src/server/handlers/nodes.rs
maschad: Good point 24c7dcf

maschad:
> You are not storing the `tx_digest` in the stacks table.

Yes my mistake @jorgeantonio21  I've updated the PR https://github.com/atoma-network/atoma-proxy/pull/424/files#diff-75032dcf8abf822e05020e641f3b69e7f70cee326e71064d979dbf0cb30257baR2009

atoma-proxy/src/server/handlers/nodes.rs
```diff
@@ -331,12 +331,22 @@ pub async fn nodes_create_lock(
...
+                        endpoint: NODES_CREATE_LOCK_PATH.to_string(),
+                    })?;
+
             let public_key = STANDARD.encode(node_public_key.public_key);
             lock_compute_units_in_memory(&state, stack_small_id, timeout, max_num_tokens);
             Ok(Json(NodesCreateLockResponse {
                 public_key,
                 node_small_id: node_public_key.node_small_id as u64,
-                stack_entry_digest: None,
+                stack_entry_digest: tx_digest,
```
jorgeantonio21, line 349: This is not needed, I believe.

atoma-state/src/state_manager.rs
```diff
@@ -2006,8 +2006,8 @@ impl AtomaState {
     ) -> Result<()> {
         sqlx::query(
             "INSERT INTO stacks
-                (owner, stack_small_id, stack_id, task_small_id, selected_node_id, num_compute_units, price_per_one_million_compute_units, already_computed_units, locked_compute_units, in_settle_period, total_hash, num_total_messages, user_id, acquired_timestamp)
-                VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14)",
+                (owner, stack_small_id, stack_id, task_small_id, selected_node_id, num_compute_units, price_per_one_million_compute_units, already_computed_units, locked_compute_units, in_settle_period, total_hash, num_total_messages, user_id, acquired_timestamp, tx_digest)
+                VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15)",
         )
```
jorgeantonio21, line 2011: You don't bind the `stack.transaction` at the end, so this will crash

jorgeantonio21: Also, the transaction digest is not present in the original on-chain event. So there is some work to be done on the event subscriber.

jorgeantonio21 (changes requested):

atoma-proxy/src/server/handlers/nodes.rs
maschad: Why is that?

atoma-state/src/tests.rs
```diff
@@ -346,6 +350,7 @@ async fn create_test_stack(
     .bind(0i64) // Default num_total_messages
     .bind(user_id)
     .bind(chrono::Utc::now()) // Acquired timestamp
+    .bind(tx_digest)
```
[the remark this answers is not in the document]
maschad: This is wrong

jorgeantonio21 (approved):

PR #427, opened 2025-04-24
feat: add tables for fiat

Cifko:
Add tables for fiat.

### DIFF
--- a/atoma-proxy-service/src/handlers/auth.rs
+++ b/atoma-proxy-service/src/handlers/auth.rs
@@ -586,7 +586,7 @@ pub async fn get_balance(
     Ok(Json(
         proxy_service_state
             .atoma_state
-            .get_balance_for_user(user_id)
+            .get_crypto_balance_for_user(user_id)
             .await
             .map_err(|e| {
                 error!("Failed to get balance: {:?}", e);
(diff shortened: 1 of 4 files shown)


jorgeantonio21 (approved):

atoma-proxy/src/server/handlers/mod.rs
```diff
@@ -82,6 +82,28 @@ pub fn update_state_manager(
     Ok(())
 }
 
+pub fn update_state_manager_fiat(
```
maschad, line 85: This function is never called, I imagine we want to perform this once the AI service request ahas been performed

atoma-proxy/src/server/middleware.rs
```diff
@@ -1962,17 +1994,54 @@ pub mod auth {
...
+                client_message: None,
+                endpoint: endpoint.to_string(),
+            })?
+            .map_err(|err| AtomaProxyError::InternalError {
+                message: format!("Failed to get LockUserFiatBalance result: {err:?}"),
+                client_message: None,
+                endpoint: endpoint.to_string(),
+            })?;
+
+        if locked_fiat {
```
maschad, line 2026: How is the user's fiat balance unlocked? wouldn't this be after they have used up all their tokens?

maschad (changes requested):
Left some comments

PR #428, opened 2025-04-25
feat: fiat requests/refund

Cifko:


atoma-proxy/src/server/handlers/nodes.rs
```diff
@@ -373,6 +373,8 @@ pub async fn nodes_create_lock(
                     stack_small_id,
                     selected_node_id,
                     tx_digest,
+                    fiat_locked_amount: _, // TODO: fiat
```
jorgeantonio21, line 376: How come todo fiat ?

atoma-proxy/src/server/handlers/nodes.rs
```diff
@@ -438,6 +440,7 @@ pub async fn nodes_create_lock(
                     })?;
                 if let Some(node_public_key) = node_public_key {
                     let public_key = STANDARD.encode(node_public_key.public_key);
+                    let stack_small_id = stack_small_id.unwrap();
```
jorgeantonio21, line 443: Don't unwrap

atoma-proxy/src/server/streamer.rs
```diff
@@ -105,9 +110,11 @@ impl Streamer {
     pub fn new(
```
jorgeantonio21, line 110: Update the docs to explicitly mention that one of stack small id or fiat estimated amount needs to be provided.

atoma-proxy/src/server/handlers/mod.rs
```diff
@@ -82,6 +82,53 @@ pub fn update_state_manager(
...
+)]
+pub fn update_state_manager_fiat(
+    state_manager_sender: &Sender<AtomaAtomaStateManagerEvent>,
+    user_id: i64,
+    estimated_amount: i64,
+    amount: i64,
+    endpoint: &str,
+) -> Result<()> {
+    // Update stack num tokens
+    state_manager_sender
```
jorgeantonio21, line 118: You need to receive the response back, to know if you actually updated any row, or not. In the latter case, some double spending was tried and this should be properly handled.

atoma-state/src/state_manager.rs
```diff
@@ -4492,6 +4492,83 @@ impl AtomaState {
...
+    pub async fn lock_user_fiat_balance(&self, user_id: i64, amount: i64) -> Result<bool> {
+        let result = sqlx::query(
+            "UPDATE fiat_balance SET overcharged_unsettled_amount = overcharged_unsettled_amount + $2 WHERE user_id = $1 AND usd_balance >= already_debited_amount + overcharged_unsettled_amount + $1",
+        )
+        .bind(user_id)
+        .bind(amount)
+        .execute(&self.db)
+        .await?;
+
+        Ok(result.rows_affected() == 1)
```
jorgeantonio21, line 4524: Are you sure it is only one or 0 rows affected ?

atoma-state/src/state_manager.rs
```diff
@@ -4492,6 +4492,83 @@ impl AtomaState {
...
+        let result = sqlx::query(
+            "UPDATE fiat_balance SET already_debited_amount = already_debited_amount + $3, overcharged_unsettled_amount = overcharged_unsettled_amount - $2 WHERE user_id = $1",
+        )
+        .bind(user_id)
+        .bind(estimated_amount)
+        .bind(amount)
+        .execute(&self.db)
+        .await?;
+
+        if result.rows_affected() == 0 {
```
jorgeantonio21, line 4566: When can this happen ? User not found ?

atoma-proxy/src/server/middleware.rs
```diff
@@ -345,36 +353,70 @@ pub async fn authenticate_middleware(
...
+        let req = match stack_small_id {
+            Some(stack_small_id) => {
+                match utils::try_validate_stack_for_request(
+                    &state,
+                    &body_json,
+                    &mut req_parts,
+                    selected_node_id,
                     stack_small_id,
-                    max_total_compute_units as i64,
-                    0,
```
jorgeantonio21, line 374: What was the 0 for ? And why don't we keep it ?

atoma-proxy/src/server/middleware.rs
```diff
@@ -345,36 +353,70 @@ pub async fn authenticate_middleware(
...
-            user_id,
-            &endpoint,
-        )
-        .await
-        {
-            Ok(req) => req,
-            Err(e) => {
-                update_state_manager(
-                    &state.state_manager_sender,
+        let req = match stack_small_id {
```
jorgeantonio21, line 364: Can we encapsulate these checks into a nice macro ? It seems we are always repeating the logic

atoma-proxy/src/server/middleware.rs
```diff
@@ -716,16 +757,15 @@ pub async fn handle_locked_stack_middleware(
...
-                &[
-                    KeyValue::new(
-                        "stack_small_id",
-                        request_metadata.selected_stack_small_id.to_string(),
-                    ),
-                    KeyValue::new("user_id", request_metadata.user_id),
-                ],
-            );
+            if let Some(stack_small_id) = request_metadata.selected_stack_small_id {
+                UNAVAILABLE_STACK_COUNTER_PER_USER.add(
```
jorgeantonio21, line 761: Should we have similar prometheus metrics for the fiat payments ?

atoma-proxy/src/server/middleware.rs
```diff
@@ -1962,17 +2036,57 @@ pub mod auth {
...
-            user_id,
-            lock_guard,
-            endpoint.to_string(),
-            total_tokens,
-            Arc::clone(&state.sui),
-            node,
-        )
-        .await
+        // We don't have a stack for the user, lets check if the user is using fiat currency.
+        let (result_sender, result_receiver) = oneshot::channel();
```
jorgeantonio21, line 2040: Encapsulate this logic into a function, this possibly needs to be spawned on a new task as well ? For no cancellations hold state ?

atoma-proxy/src/server/middleware.rs
```diff
@@ -2211,7 +2327,136 @@ pub mod utils {
...
+        price_per_million: i64,
+        num_input_tokens: u64,
+        total_compute_units: u64,
+        user_id: i64,
+        endpoint: &str,
+    ) -> Result<Request<Body>> {
+        let ProcessedRequest {
+            node_address,
+            signature,
+        } = auth::process_selected_stack(state, body_json, selected_node_id, endpoint).await?;
```
jorgeantonio21, line 2423: Why ? This is not for stack based requests.

atoma-proxy/src/server/middleware.rs
```diff
@@ -2211,7 +2327,136 @@ pub mod utils {
...
+    /// * Stack ID
+    /// * Endpoint path
+    /// * Model name
+    #[instrument(level = "info", skip_all, fields(
+        %endpoint,
+        %fiat_estimated_amount,
+        %user_id
+    ), err)]
+    #[allow(clippy::too_many_arguments)]
+    pub async fn create_fiat_request(
```
jorgeantonio21, line 2408: Do we want to avoid replay attacks when these come not from the proxy ?

jorgeantonio21 (changes requested):
Looking good but left a bunch of important comments to be addressed.

atoma-proxy/src/server/handlers/mod.rs
Cifko: This function is used only to do refunds, when the computation is done or cancelled. It's not used for locking funds.

atoma-proxy/src/server/middleware.rs
Cifko: It's now couple lines before. This is full refund, We charge 0 tokens/usd in case of failure on our side.

atoma-proxy/src/server/middleware.rs
Cifko: Since fiat is post payment (it's always accepted by the node) you will never receive TOO_EARLY (this is when the stack was not found on the node. In case of fiat this will never happen, either you are in the white list or not.

atoma-proxy/src/server/middleware.rs
Cifko: This is already inside tokio task called from authenticate_middleware.

atoma-proxy/src/server/middleware.rs
Cifko: Yes, don't know how yet. But how would an attacker get hands on such request? It's not public.

atoma-proxy/src/server/handlers/nodes.rs
Cifko: working on it.

atoma-proxy/src/server/middleware.rs
Cifko: Yeah, we do refund a lot, I will take a look

atoma-proxy/src/server/handlers/nodes.rs
Cifko: I will add the confidential in different PR

jorgeantonio21 (approved):

PR #432, opened 2025-04-28
fix: infinite stream kill

Cifko:
Fix the infinite stream kill.


jorgeantonio21 (approved):

jorgeantonio21 (changes requested):
This is not intended, we should immediately return too many requests to the client.

PR #444, opened 2025-04-29
feat: add metrics to track latency from proxy to node

maschad:
We would like to see what the network bandwidth latency between the nodes and the proxy is to ensure that this is not negatively impacting service requests

### DIFF
--- a/atoma-state/src/network.rs
+++ b/atoma-state/src/network.rs
@@ -6,10 +6,20 @@ use opentelemetry::{
 use reqwest::Client;
 use std::sync::LazyLock;
 use std::time::{Duration, Instant};
-use tracing::{debug, error, warn};
+use tracing::{debug, error, instrument, warn};
 
 static GLOBAL_METER: LazyLock<Meter> = LazyLock::new(|| global::meter("atoma-proxy"));
 
+/// Histogram metric that tracks the latency of node connections from the proxy to the node.
+///
+/// This metric tracks the latency of node connections from the proxy to the node.
+///
+/// # Metric Details
+/// - Name: `atoma_node_connection_latency`
+/// - Type: Histogram
+/// - Labels: `node_ip_address`
+/// - Unit: seconds (s)
+///
 pub static NODE_CONNECTION_LATENCY: LazyLock<Histogram<f64>> = LazyLock::new(|| {
     GLOBAL_METER
         .f64_histogram("atoma_node_connection_latency")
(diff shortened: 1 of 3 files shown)


atoma-state/src/state_manager.rs
```diff
@@ -135,8 +136,25 @@ impl AtomaStateManager {
     /// ```
     #[instrument(level = "trace", skip_all)]
     pub async fn run(self, mut shutdown_signal: Receiver<bool>) -> Result<()> {
+        let mut network_metrics = NetworkMetrics::new();
+        let interval = std::time::Duration::from_secs(15);
+
         loop {
             tokio::select! {
+                 () = tokio::time::sleep(interval) => {
```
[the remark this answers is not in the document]
maschad: Good point! addressed in cc0bd729fa7e64c24618057ba3454147de492e9c

atoma-state/src/network.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl NetworkMetrics {
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            client: Client::new(),
+        }
+    }
+
+    pub async fn update_metrics(&mut self, node_addresses: Vec<String>) {
```
jorgeantonio21, line 42: Add docs and instrumentation

atoma-state/src/network.rs
```diff
@@ -0,0 +1,90 @@
...
+        for address in node_addresses {
+            let url = format!("http://{address}:3000/health");
```
jorgeantonio21, line 44: Why port 3000 ?

atoma-state/src/state_manager.rs
```diff
@@ -135,8 +136,38 @@ impl AtomaStateManager {
...
     #[instrument(level = "trace", skip_all)]
     pub async fn run(self, mut shutdown_signal: Receiver<bool>) -> Result<()> {
+        let mut network_metrics = NetworkMetrics::new();
+        let interval = std::time::Duration::from_secs(15);
+
+        // Create a channel for interval-based updates
+        let (interval_tx, interval_rx) = flume::unbounded();
+
+        // Spawn a task that sends a message every interval
+        tokio::spawn(async move {
```
jorgeantonio21, line 146: Why spawning a new task for this ? The atoma state manager is already spawned on a new task, in the main binary

jorgeantonio21 (commented):
Looking good, left a few comments

atoma-state/src/state_manager.rs
maschad: As mentioned [here](https://github.com/atoma-network/atoma-proxy/pull/444#discussion_r2068983119) having it in the same thread wouldn't work thus I created a separate thread. This shouldn't affect the other thread regardless so it should be fine

atoma-state/src/network.rs
maschad: Good point :) f4ce906

atoma-state/src/network.rs
maschad: Good catch!

jorgeantonio21 (approved):

PR #453, opened 2025-05-06
feat: add model per intentional cancellations metrics

jorgeantonio21:


atoma-state/src/state_manager.rs
```diff
@@ -183,7 +183,6 @@ impl AtomaStateManager {
                                     error = %e,
                                     "Error handling Atoma event"
                                 );
-                                continue;
```
maschad, line 186: Should these be included on this branch?

maschad (changes requested):
Looks good but I think you accidentally included some unrelated changes

atoma-state/src/state_manager.rs
jorgeantonio21: That's totally fine

PR #456, opened 2025-05-07
feat: add node exporter for storage usage observability

maschad:
This will help to give us observability over our disk usage and other OS metrics needed for alerts.

In this case I can monitor `node_filesystem_avail_bytes{mountpoint="/var/lib/docker"}`
Screenshot 2025-05-07 at 12 15 06 PM

where we could have alerts such as 

```
(
  100 * (1 - node_filesystem_avail_bytes{mountpoint="/"} / node_filesystem_size_bytes{mountpoint="/"}) > 85
  and
  rate(node_disk_read_bytes_total[5m]) > 0
)
```

when our disk usage exceeds 85% for instance

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -218,6 +218,24 @@ services:
       - atoma-network
     user: "0"
 
+  node-exporter:
+    image: prom/node-exporter:v1.7.0
+    container_name: node-exporter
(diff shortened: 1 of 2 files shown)


jorgeantonio21 (approved):
Nice

PR #457, opened 2025-05-07
fix: optimize total hash storage

maschad:
This PR drops the unused `total_hash` field which is no longer needed since we are no longer relying on cross validation sampling consensus


jorgeantonio21 (approved):
Great job on this

PR #461, opened 2025-05-09
feat: completions endpoint refactor

jorgeantonio21:


atoma-proxy/src/server/handlers/completions.rs
```diff
@@ -0,0 +1,1311 @@
...
+            .and_then(serde_json::Value::as_bool)
+            .unwrap_or_default();
+
+        match handle_completions_request(&state, &metadata, headers, payload, is_streaming).await {
+            Ok(response) => {
+                if !is_streaming {
+                    // The streaming metric is recorded in the streamer (final chunk)
+                    TOTAL_COMPLETED_REQUESTS.add(1, &[KeyValue::new("model", metadata.model_name)]);
+                }
+                Ok(response)
```
Cifko, line 133: Update stack/Fiat. So the amount is unlocked and counted as completed.

atoma-proxy/src/server/handlers/completions.rs
```diff
@@ -0,0 +1,1311 @@
...
```
Cifko, line 396: Same here. Update the stack/fiat.

Cifko (approved):

atoma-proxy/src/server/handlers/completions.rs
jorgeantonio21: This is done within the function body for non-streaming requests, and handle final chunk or at drop for streaming requests.

atoma-proxy/src/server/handlers/completions.rs
jorgeantonio21: See above

PR #471, opened 2025-05-13
fix: update fiat

Cifko:
Rework the fiat payment to the same way it's done on the node. This also fixes non-streaming chat. Update the `usage_per_model` table.


atoma-proxy/src/server/streamer.rs
```diff
@@ -618,9 +615,10 @@ impl Drop for Streamer {
                 if let Err(e) = update_state_manager_fiat(
                     &self.state_manager_sender,
                     self.user_id,
-                    self.fiat_estimated_amount.unwrap_or_default(),
-                    self.num_generated_tokens * self.price_per_million.unwrap_or_default()
-                        / ONE_MILLION as i64,
+                    self.estimated_total_tokens,
+                    self.num_generated_tokens,
+                    self.price_per_million.unwrap_or_default(),
```
jorgeantonio21, line 620: unwrap or default ? How come ? Default will be 0..

atoma-proxy/src/server/handlers/completions.rs
```diff
@@ -151,8 +151,10 @@ pub async fn completions_create(
                     update_state_manager_fiat(
                         &state.state_manager_sender,
                         metadata.user_id,
-                        metadata.fiat_estimated_amount.unwrap_or_default(),
+                        metadata.max_total_num_compute_units as i64,
                         0,
+                        metadata.price_per_million.unwrap_or_default(),
```
jorgeantonio21, line 156: How come unwrap or default ? The price should never be 0, I believe..

jorgeantonio21 (commented):
Looks good, left a few comments though

atoma-proxy/src/server/streamer.rs
Cifko: ok, now the price per million always filled

atoma-proxy/src/server/handlers/completions.rs
Cifko: same as above

jorgeantonio21 (approved):

PR #486, opened 2025-05-26
feat: increase max pgpool connections

jorgeantonio21:


maschad (commented):
What's the rationale behind this PR?

atoma-state/src/metrics.rs
```diff
@@ -318,12 +318,12 @@ pub struct NodeMetricsCollector {
     #[allow(dead_code)]
     registry: Registry,
 
+    /// Average waiting queue duration for chat completions
```
Cifko, line 321: Why is this change in this PR?

Cifko (approved):

PR #499, opened 2025-05-30
chore: refactor sending atoma state manager event

Cifko:
Refactor the sending of atoma state manager events into simple functions. It's better for code readability.


jorgeantonio21 (approved):
Nice

maschad (approved):

